PR #99, opened 2024-05-03
update ssa.Slice

visualfc:


internal/runtime/z_slice.go
```diff
@@ -54,4 +54,22 @@ func SliceData(s Slice) unsafe.Pointer {
...
+}
+
+func NewSlice3(base unsafe.Pointer, eltSize, cap, i, j, k int) (s Slice) {
+	if i < 0 || j < i || k < j || k > cap {
+		panic("slice index out of bounds")
+	}
+	s.len = j - i
+	s.cap = k - i
+	if k-i > 0 {
+		s.data = add(base, uintptr(i*eltSize), "i < k <= cap")
```
xushiwei: use c.Advance(ptr, offset)

internal/runtime/z_string.go
```diff
@@ -80,4 +80,14 @@ func CStrDup(s String) *int8 {
 	return CStrCopy(dest, s)
 }
 
+func NewStringSlice(base String, i, j int) String {
+	if i < 0 || j < i || j > base.len {
+		panic("string slice index out of bounds")
+	}
+	if i < base.len {
```
xushiwei, line 87: this `if statement` is not need.

visualfc: ```"hello"[5:]``` return empty string

PR #183, opened 2024-05-15
ssa: builtin print/println

visualfc:


ssa/expr.go
```diff
@@ -1386,6 +1386,42 @@ func (b Builder) BuiltinCall(fn string, args ...Expr) (ret Expr) {
 				}
 			}
 		}
+	case "print", "println":
+		ln := fn == "println"
+		ret.Type = b.Prog.Void()
+		for i, arg := range args {
+			if ln && i > 0 {
+				b.InlineCall(b.Func.Pkg.rtFunc("printsp"), arg)
```
xushiwei: Don't call private func. We will use all public symbols to define runtime spec later.

visualfc: fixed and export print symbols

PR #246, opened 2024-05-29
ssa: update linkname go style

visualfc:
```
// likname:
// - func: pkg.name
// - method: pkg.T.name 
// - method: pkg.(*T).name

// local in pkg name
// - func: name
// - method: T.name 
// - method: (*T).name
```


ssa/type.go
```diff
@@ -479,16 +479,17 @@ func PathOf(pkg *types.Package) string {
 
 // FuncName:
 // - func: pkg.name
-// - method: (pkg.T).name, (*pkg.T).name
+// - method: pkg.(T).name, pkg.(*T).name
```
xushiwei: pkg.T.name

PR #298, opened 2024-06-11
ssa: interface equal

visualfc:


ssa/type_cvt.go
```diff
@@ -128,6 +128,7 @@ func (p goTypes) cvtNamed(t *types.Named) (raw *types.Named, cvt bool) {
 		named.SetUnderlying(tund)
 		return named, true
 	}
+	p.typs[unsafe.Pointer(t)] = unsafe.Pointer(t)
```
xushiwei, line 131: move `p.typs[unsafe.Pointer(t)] = unsafe.Pointer(named)` into `if` statement.

internal/runtime/z_face.go
```diff
@@ -322,4 +325,95 @@ func Implements(T, V *abi.Type) bool {
 	return false
 }
 
+func EfaceEqual(v, u eface) bool {
```
xushiwei, line 328: https://github.com/goplus/llgo/blob/main/internal/abi/type.go#L53

How to use `abi.Type.Equal`?

ssa/type_cvt.go
visualfc: 提前使用 `p.typs[unsafe.Pointer(t)] = unsafe.Pointer(named)` 是为了避免循环定义的问题，所以还需要在未使用时恢复。

PR #337, opened 2024-06-17
readme:lld-17 download on linux

### LINKED ISSUE #336: clang: error: invalid linker name in argument '-fuse-ld=lld'r on Linux/Debian
```bash
root@ef137ede8eb2:~/llgobuild/_demo/concat# llgo run .
clang: error: invalid linker name in argument '-fuse-ld=lld'
```
lld-17 needs to be installed, and it can run normally
```
 apt-get install lld-17
```

luoliwoshang:
Fix #336
brew installs LLVM with its own lld
```bash
brew install llvm@17
❯ which ld.lld 
/opt/homebrew/opt/llvm@17/bin/ld.lld
```
However, `apt-get install --no-install-recommends llvm-17-dev` does not carry this lld，additional downloads are required, Otherwise, the #336 problem will occur

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,7 @@ wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
 sudo apt-get update  # execute if needed
 sudo apt-get install libgc-dev
 sudo apt-get install --no-install-recommends llvm-17-dev
+sudo apt-get install lld-17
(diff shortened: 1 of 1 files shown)


luoliwoshang:
Already solved in #338

luoliwoshang:
Already solved in #338

PR #363, opened 2024-06-19
runtime: tracePanic

visualfc:
TODO: panic kind to string
`println("(", e.Kind(), ")", e.data)`


c/bitcast/_cast/cast.c
```diff
@@ -0,0 +1,17 @@
+typedef union {
+    double d;
+    float  f;
+    long   v;
+} castUnion;
+
+double bitcastToFloat64(long v) {
```
xushiwei: use `llgo`  namespace. eg. `llgoToFloat64`

c/bitcast/bitcast.go
```diff
@@ -0,0 +1,14 @@
+package bitcast
```
xushiwei, line 17: add copyright

PR #402, opened 2024-06-24
add c/math/rand

tsingbx:
add math/rand


.vscode/launch.json
```diff
@@ -0,0 +1,35 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
xushiwei, line 2: don't commit .vscode config into git

c/math/rand/rand.go
```diff
@@ -0,0 +1,23 @@
...
+
+const (
+	LLGoPackage = "decl"
+)
+
+//go:linkname Rand C.rand
+func Rand() c.Int
+
+//go:linkname Rand_r C.rand_r
+func Rand_r(*c.Uint) c.Int
```
xushiwei: Rand_r => RandR

PR #403, opened 2024-06-24
add math/rand 

tsingbx:
add math/rand


luoliwoshang:
Maybe we can merge some unnecessary commit records.

PR #430, opened 2024-06-27
[WIP] llgo/c/lua

luoliwoshang:
- [ ] readme
- [x] test in mac
- [ ] test in linux

#### basic
- [x] stack operate
- [x] func call 
- [ ] userdata 
- [x] thread
- [ ] closure
#### table
- [x] table
- [ ] metatable
- [ ] metamethod 
#### corroutine
- [x] corroutine-resume
- [ ] continuation-function
- [ ] corroutine-yield

## problem
- [x] Can not start the corroutine normally
 
**debug**
1. The number of values returned when the nres is not assigned after the resume is executed
2. After executing lua.resume, the function within the coroutine should be popped from the stack, and the return values of the corresponding function should be pushed onto the stack. In this example, the coroutine's function was not popped from the stack, and the return values were not pushed onto the stack.
```c
func printStack(L *lua.Lua_State, msg *c.Char) {
	top := L.GetTop()
	c.Printf(c.Str("%s\nStack size: %d\n"), msg, top)
	for i := 1; i <= int(top); i++ {
		t := L.Type(c.Int(i))
		c.Printf(c.Str("%d: %s - "), i, L.TypeName(t))
		switch t {
		case c.Int(lua.NUMBER):
			// Assuming all numbers are floats for safety in printing
			c.Printf(c.Str("%g\n"), L.ToNumber(c.Int(i)))
		case c.Int(lua.STRING):
			c.Printf(c.Str("%s\n"), L.ToString(c.Int(i)))
		case c.Int(lua.BOOLEAN):
			c.Printf(c.Str("%t\n"), L.ToBoolean(c.Int(i)))
		default:
			c.Printf(c.Str("other\n"))
		}
	}
}

func main() {
	L := lua.NewState()
	defer L.Close()
	L.OpenLibs()

	if L.DoString(c.Str("coro_func = function() for i=1,5 do coroutine.yield(i) end end")) != lua.OK {
		println("load err")
	}

	L.GetGlobal(c.Str("coro_func"))

	co := L.NewThread()
	L.PushValue(-2) // copy function to head
	L.XMove(co, 1)  // move function to coroutine

	printStack(L, c.Str("main stack info"))
	printStack(co, c.Str("corroutine stack info"))
	println("----------")

	var nres c.Int
	status := co.Resume(nil, 0, &nres) // start coroutine

	printStack(L, c.Str("After resume main stack info"))
	// Stack size: 2
	// 1: function - other
	// 2: thread - other
	printStack(co, c.Str("After corroutine stack info"))

	// Stack size: 1
	// 1: function - other

	// After executing lua.resume, the function within the coroutine should be popped from the stack, and the return values of the corresponding function should be pushed onto the stack.
	// In this example, the coroutine's function was not popped from the stack, and the return values were not pushed onto the stack.

	if status == lua.YIELD || status == lua.OK {
		c.Printf(c.Str("Status %d Coroutine returned %d results.\n"), status, nres) // Status 0 Coroutine returned 0 results.
		// Nres should be 1, which means holding a return value, and status should be lua.YIELD.
		// But in the test, the nres value is 0, and the status is lua.OK.
		for i := c.Int(1); i <= nres; i++ {
			if co.IsNumber(-i) != 0 {
				c.Printf(c.Str("Result %d: %g\n"), i, co.ToNumber(-i))
			} else {
				c.Printf(c.Str("Other type\n"))
			}
		}
	}
}
```
---

- [x] When a parameter of Integer is pushed and called, the program gets stuck

The reason comes from a llgo:link writing error:
It should be llgo:link (* Lua_State). PushInteger C. Lua _ pushinteger.
And I wrote llgo:link (* Lua_Stage). PushInteger C. Lua _ pushinteger

llgo:link should warn when method not found  #435 
```go
func main() {
	L := lua.Newstate()
	defer L.Close(
	L.Openlibs()
	code := c.Str(`function combineParams(int, num, str)
	return 'Combined Result: ' .. str .. ' ' .. tostring(int) .. ' ' .. tostring(num)
end`)
	if res := L.Dostring(code); res != lua.OK {
		c.Printf(c.Str("error: %s\n"), L.Tostring(-1))
	}
	L.Getglobal(c.Str("combineParams"))
	L.Pushinteger(42)
	L.Pushnumber(3.14159)
	L.Pushstring(c.Str("Hello, World!"))
	if res := L.Pcall(3, 1, 0); res != lua.OK {
		c.Printf(c.Str("error: %s\n"), L.Tostring(-1))
	}
	if res := L.Isstring(-1); res != 0 {
		result := L.Tostring(-1)
		c.Printf(result)
	}
}
```


c/lua/lua.go
```diff
@@ -0,0 +1,495 @@
...
+const (
+	OK        = 0
+	YIELD     = 1
+	ERRRUN    = 2
+	ERRSYNTAX = 3
+	ERRMEM    = 4
+	ERRERR    = 5
+)
+
+type Lua_State struct {
```
xushiwei: `State` is better.

c/lua/lua.go
```diff
@@ -0,0 +1,495 @@
...
+
+// /* predefined values in the registry */
+const (
+	RIDX_MAINTHREAD = 1
+	RIDX_GLOBALS    = 2
+	RIDX_LAST       = RIDX_GLOBALS
+)
+
+// /* type of numbers in Lua */
+type Number c.Double
```
xushiwei: type alias is better

c/lua/lua.go
```diff
@@ -0,0 +1,495 @@
...
+
+// /* type for integer functions */
+// TODO(zzy):consider dynamic size
+
+type Integer = c.Int
+
+// /* unsigned integer type */
+type Unsigned c.Uint
```
xushiwei: type alias is better

PR #431, opened 2024-06-28
ssa: support runtime.map

visualfc:
**ssa/runtime**
- ssa: index takeAddress
- ssa: runtime type.equal
- runtime: check abi.KindDirectIface
- runtime: check abi.TFlagRegularMemory

**support runtime.map**
- ssa.MakeMap
- ssa.MapUpdate
- ssa.Lookup
- ssa.Range
- ssa.Next
- map eql nil
- map delete
- map clear ( Go1.21 ~ )
- mapType.hasher


internal/runtime/map.go
```diff
@@ -674,12 +674,15 @@ bucketloop:
 		kmem := newobject(t.Key)
 		*(*unsafe.Pointer)(insertk) = kmem
 		insertk = kmem
+		*(*unsafe.Pointer)(insertk) = key
```
xushiwei: why need this?

visualfc: 测试代码，已删除。

PR #436, opened 2024-07-02
link:not exist method warning

### LINKED ISSUE #435: llgo:link should warn when method not found
The expected behavior for llgo:link should be as follows:

```go
// llgo:link (*Lua_State).PushInteger C.lua_pushinteger
func (L *Lua_State) PushInteger(n Integer) {}
```
However, after writing the following content:
```go
//Should be Lua_State not Lua_Stage

// llgo:link (*Lua_Stage).PushInteger C.lua_pushinteger
func (L *Lua_State) PushInteger(n Integer) {}
```
there is no warning for a failed link, leading to runtime errors in the program.

luoliwoshang:
fix #435 
Removes the check that only throws an error when a function does not exist.

When using the erroneous example from #435, a warning message will be displayed:
```go
// Should be Lua_State, not Lua_Stage

// llgo:link (*Lua_Stage).PushInteger C.lua_pushinteger
func (L *Lua_State) PushInteger(n Integer) {}
```
Running the code will result in the following warning being correctly output:
```bash
❯ llgo run .
==> // llgo:link (*Lua_Stage).PushInteger C.lua_pushinteger
llgo: linkname (*Lua_Stage).PushInteger not found and ignored
```

### DIFF
--- a/cl/import.go
+++ b/cl/import.go
@@ -270,7 +270,7 @@ func (p *context) initLink(line string, prefix int, f func(inPkgName string) (fu
 			} else {
 				panic(line + ": no specified call convention. eg. //go:linkname Printf C.printf")
 			}
-		} else if c := inPkgName[0]; c >= 'A' && c <= 'Z' {
+		} else {
 			fmt.Fprintln(os.Stderr, "==>", line)
 			fmt.Fprintf(os.Stderr, "llgo: linkname %s not found and ignored\n", inPkgName)
 		}


cl/import.go
```diff
@@ -270,13 +271,19 @@ func (p *context) initLink(line string, prefix int, f func(inPkgName string) (fu
 			} else {
 				panic(line + ": no specified call convention. eg. //go:linkname Printf C.printf")
 			}
-		} else if c := inPkgName[0]; c >= 'A' && c <= 'Z' {
+		} else if isExported(inPkgName) {
```
xushiwei: maybe just remove `if cond`.

luoliwoshang: Done，I mistakenly thought there was some consideration for non-exported functions here.

PR #444, opened 2024-07-03
[wip] llgo/rust:demo

luoliwoshang:
### link dylib
1. compile rust to dylib
`lib.rs`
```rust
#[no_mangle]
pub extern "C" fn sort(arr: *mut i32, len: usize) {
    unsafe {
        if arr.is_null() {
            return;
        }
        let slice = std::slice::from_raw_parts_mut(arr, len);
        slice.sort_unstable();
    }
}
```
`Cargo.toml`
```toml
[lib]
name = "example_lib"
crate-type = ["cdylib"]
```
`cbindgen --config cbindgen.toml --crate example_lib --output example_lib.h`
```c
#ifndef EXAMPLE_LIB_H
#define EXAMPLE_LIB_H

#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

void sort(int32_t *arr, uintptr_t len);

#endif /* EXAMPLE_LIB_H */
```
3. llgo link `libexample_lib.dylib`
```go
const (
	LLGoPackage = "link: -L/Users/xxxx/Documents/Code/rust/example_lib/ctest -lexample_lib"
)
```
4. Write the .pc file for the dylib generated by the Rust library and copy it to pkg-config for proper location.
`sudo cp example_lib.pc /opt/homebrew/lib/pkgconfig`
```bash
prefix=/Users/xxxxxxDocuments/Code/rust/example_lib
exec_prefix=${prefix}
libdir=${exec_prefix}/target/release

Name: example_lib
Description: An example Rust library
Version: 1.0
Libs: -L${libdir} -lexample_lib
```
5. can use pkg-config to find the dynamic library.
```
❯ pkg-config --libs example_lib                     
-L/Users/xxxx/Documents/Code/rust/example_lib/target/release -lexample_lib
```
```
const (
	LLGoPackage = "link: $(pkg-config --libs example_lib); -lexample_lib"
)
```
6. Write a Demo based on the generated header file,can run correctly
```go
package rust

import "github.com/goplus/llgo/c"

const (
	LLGoPackage = "link: $(pkg-config --libs example_lib); -lexample_lib"
)

//llgo:link Sort C.sort
func Sort(arr *c.Int, len c.Ulong) {}
```
```go
func main() {
	vals := []c.Int{32, 58, 25, 92, 45, 78}
	rust.Sort(unsafe.SliceData(vals), c.Ulong(len(vals)))
	for _, v := range vals {
		println(v)
	}
}
```
image

#### link llvm ir
1. Compile the functions exported by Rust as a library into LLVM IR.
3. Link the LLVM IR representation of this library during the initialization of initMainPkg.
7. Specify Rust's dynamic libraries during compilation with clang.
8. At runtime, specify the environment variable using: `export DYLD_LIBRARY_PATH=/Users/zhangzhiyang/.rustup/toolchains/stable-aarch64-apple-darwin/lib:$DYLD_LIBRARY_PATH.`

image

**Issues to consider/discuss**
1. rust's standard library does not support PGG-config for manual search, and the location of the version of rust library needs to be specified manually. Moreover, the installed dynamic link inventory is in different hashes, and additional processing needs to be done for links during rust compilation.
3. Check whether the link library can be used normally after it is compiled into a dynamic or static link library
4. How to find links for Rust libraries compiled as Lib libraries


xushiwei:
Import frequently used rust packages instead of demo

PR #449, opened 2024-07-05
feat(rust): implement rust std hashmap

hackerchai:
See [hashmap-rs-ffi-c](https://github.com/hackerchai/hashmap-rs-ffi-c) for install rust dylib.


xushiwei:
rust's hashmap has no meaning to llgo.

PR #450, opened 2024-07-05
feat(rust): implement rust sled demo

hackerchai:
Signed-off-by: hackerchai <i@hackerchai.com>


xushiwei:
please write a README file like this: https://github.com/goplus/llgo/blob/main/c/sqlite/README.md

rust/sled/sled.go
```diff
@@ -0,0 +1,35 @@
+package sled
+
+import "github.com/goplus/llgo/c"
+
+// Write the .pc file for the dylib generated by the Rust library and copy it to pkg-config for proper location.
+const (
+	LLGoPackage = "link: $(pkg-config --libs sled); -lsled"
+)
+
+type SledConfig struct {
```
xushiwei: change `SledConfig` => `Config`

rust/sled/sled.go
```diff
@@ -0,0 +1,35 @@
...
+	Unused [8]byte
```
xushiwei, line 11: should have detail members (to get correct struct size) since we call `conf := &sled.SledConfig{}` in demo.

rust/sled/sled.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type SledDb struct {
```
xushiwei: change `SledDb` => `DB`

rust/sled/sled.go
```diff
@@ -0,0 +1,35 @@
...
+	Unused [8]byte
+}
+
+//llgo:link (*SledConfig).SetPath C.sled_config_set_path
+func (conf *SledConfig) SetPath(char *c.Char) *SledConfig {
+	return nil
+}
+
+//llgo:link (*SledConfig).OpenDB C.sled_open_db
+func (conf *SledConfig) OpenDB() *SledDb {
```
xushiwei: It's not common to add OpenDB in a `Config` type.

```go
//go:linkname Open C.sled_open_db
func Open(conf *Config) *DB
```

To be confirmed: how `sled_open_db` indicates an open error?

luoliwoshang: > It's not common to add OpenDB in a `Config` type.
> 
> ```go
> //go:linkname Open C.sled_open_db
> func Open(conf *Config) *DB
> ```
> 
> To be confirmed: how `sled_open_db` indicates an open error?

In the implementation of the C wrapper in Sled, it directly calls unwrap to throw an error when there is an open error. For this C wrapper implementation, it indeed doesn't feel elegant. I think returning a null pointer for an open error might be a more elegant handling method.
```rust
/// Open a sled lock-free log-structured tree. Consumes the passed-in config.
#[no_mangle]
pub unsafe extern "C" fn sled_open_db(config: *mut Config) -> *mut Db {
    let config = Box::from_raw(config);
    Box::into_raw(Box::new(config.open().unwrap()))
}
```
When I call this function in C and pass in a null pointer, it directly throws an error:
Full Demo [https://github.com/luoliwoshang/RustTest/blob/main/sled/test/test.c](url)
```c
 printf("Opening database...\n");

free(config); // mock error

void *db = sled_open_db(config);
printf("Database opened: %p\n", db);
if (db == NULL)
{
    printf("Failed to open database\n");
    sled_free_config(config);
    return 1;
}
```
```bash
./test
Creating config...
Config created: 0x6000031e4010
Setting path...
Path set, new config: 0x6000031e4020
Opening database...
[1]    6955 segmentation fault  ./test
```

rust/sled/sled.go
luoliwoshang: In my latest [update](https://github.com/hackerchai/llgo/pull/2), the configuration is created using sled_create_config. The current status of the PR seems to have reverted this commit due to a rebase operation.

PR #453, opened 2024-07-06
[wip] ssa: select

visualfc:
**TODO**

recv is block
```
package main

func main() {
	c1 := make(chan string)
	c2 := make(chan string)
	go func() {
		c1 <- "ch1"
	}()
	go func() {
		c2 <- "ch2"
	}()

	for i := 0; i < 2; i++ {
		select {
		case msg1 := <-c1:
			println(msg1)
		case msg2 := <-c2:
			println(msg2)
		}
	}
}

```


ssa/datastruct.go
```diff
@@ -681,4 +681,102 @@ func (b Builder) Recv(ch Expr, commaOk bool) (ret Expr) {
...
+// Example printed form:
+//
+//	t3 = select nonblocking [<-t0, t1<-t2]
+//	t4 = select blocking []
+type Select struct {
+	States   []*SelectState
+	Blocking bool
+}
+
+func (b Builder) Select(blocking bool, states []*SelectState) (ret Expr) {
```
xushiwei: `blocking` as second parameters

visualfc: fixed

PR #493, opened 2024-07-11
fix:correct `Sizeof` to align size properly

### LINKED ISSUE #492: Inconsistent Structure Size from unsafe.Sizeof in LLGO Compared to Go
In the Go implementation, the size of the structure is `14`, with normal `2-byte` alignment. 
In LLGO, the size of the structure is `13`, and it does not adhere to the `2-byte` alignment typical for structures.
```go
package main

import (
    "fmt"
    "unsafe"
)

type Sockaddr_in struct {
    Sin_len    uint8
    Sin_family uint8
    Sin_port   uint16
    Sin_addr   In_addr
    Sin_zero   [8]int8
}

type In_addr struct {
    S_addr uint8
}

func main() {
    fmt.Println(unsafe.Sizeof(Sockaddr_in{}))
}
```

luoliwoshang:
fix #492 

- [x] add tests for struct size calculation

### DIFF
--- a/ssa/type.go
+++ b/ssa/type.go
@@ -114,7 +114,7 @@ func (p *goProgram) Offsetsof(fields []*types.Var) (ret []int64) {
 func (p *goProgram) Sizeof(T types.Type) int64 {
 	prog := Program(p)
 	ptrSize := int64(prog.PointerSize())
-	return prog.sizes.Sizeof(T) + extraSize(T, ptrSize)
+	return align((prog.sizes.Sizeof(T) + extraSize(T, ptrSize)), prog.sizes.Alignof(T))
 }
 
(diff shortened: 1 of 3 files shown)


ssa/type.go
```diff
@@ -114,7 +114,7 @@ func (p *goProgram) Offsetsof(fields []*types.Var) (ret []int64) {
 func (p *goProgram) Sizeof(T types.Type) int64 {
 	prog := Program(p)
 	ptrSize := int64(prog.PointerSize())
-	return prog.sizes.Sizeof(T) + extraSize(T, ptrSize)
+	return align((prog.sizes.Sizeof(T) + extraSize(T, ptrSize)), prog.sizes.Alignof(T))
```
visualfc: 只需要对结构体做修正。

PR #502, opened 2024-07-12
feat(c/os): add fcntl

spongehah:
Don't know why there is no output on line 50 of the demo and beyond

(base) spongehah@MacBook fcntl % llgo run .
Written 13 bytes to testfile.txt
File flags: 65537
set file status successfully

There is no output after that.


c/fcntl/fcntl.go
```diff
@@ -0,0 +1,42 @@
+package fcntl
```
xushiwei: merge this package into `c/os`

spongehah: image

image

I have a doubt, there is already an Open function in the os, but whether the two Opens can be equivalent

c/fcntl/fcntl.go
```diff
@@ -0,0 +1,42 @@
...
+	/* no delay */
+	O_NONBLOCK = 0x00000004
+	/* create if nonexistant */
+	O_CREAT = 0x00000200
+	/* truncate to zero length */
+	O_TRUNC = 0x00000400
+)
+
+//go:linkname FcNtl C.fcntl
+func FcNtl(a c.Int, b c.Int, vars ...any) c.Int
```
xushiwei: rename `FcNtl` => `Fcntl`

_demo/fcntl/fcntl.go
```diff
@@ -0,0 +1,77 @@
...
+	bytesRead := os.Read(fd, c.Pointer(unsafe.SliceData(buffer[:])), unsafe.Sizeof(buffer)-1)
+	if bytesRead == -1 {
+		c.Printf(c.Str("read error\n"))
+		os.Close(fd)
+		return
+	}
+
+	// Ensure that the buffer is null-terminated
+	buffer[bytesRead] = c.Char(0)
+	c.Printf(c.Str("Read %ld bytes: %s\n"), bytesRead, buffer)
```
luoliwoshang: Regarding the issue you described, the bug occurs because you are passing the array itself rather than the address of the first element when outputting the string.
```suggestion
	c.Printf(c.Str("Read %ld bytes: %s\n"), bytesRead, &buffer[0])
```

luoliwoshang (changes requested):

PR #507, opened 2024-07-13
llgo/c/lua:link style

luoliwoshang:
For macros, they are also considered to be generated by automated tools, and the naming convention is consistent with that of general functions.
```c
#define lua_tonumber(L,i)	lua_tonumberx(L,(i),NULL)
#define lua_tointeger(L,i)	lua_tointegerx(L,(i),NULL)
#define lua_pop(L,n)		lua_settop(L, -(n)-1)
```

### DIFF
--- a/c/lua/_demo/coroutine/coroutine.go
+++ b/c/lua/_demo/coroutine/coroutine.go
@@ -6,28 +6,28 @@ import (
 )
 
 func coroutineFunc(L *lua.State) {
-	L.LoadString(c.Str(`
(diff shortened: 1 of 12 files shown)


luoliwoshang:
Do automated tools expect to process macros into corresponding Go functions, or is it expected that the tool only handles Link

xushiwei:
> Do automated tools expect to process macros into corresponding Go functions, or is it expected that the tool only handles Link?

Only handle abi (binary) by default. But macro will support some special cases (eg. enum).

PR #522, opened 2024-07-16
build(macOS): change full library paths to @rpath

aofei:
```sh
$ otool -L hello
hello:
        @rpath/libgc.1.dylib (compatibility version 7.0.0, current version 7.3.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.100.2)
$ otool -l hello | grep LC_RPATH -A2
          cmd LC_RPATH
      cmdsize 32
         path @loader_path (offset 12)
--
          cmd LC_RPATH
      cmdsize 32
         path @loader_path/../lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 56
         path /opt/homebrew/Cellar/bdw-gc/8.2.6/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path /opt/homebrew/opt/bdw-gc/lib (offset 12)
```

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -17,6 +17,7 @@
 package build
 
 import (
+	"debug/macho"
 	"fmt"
 	"go/constant"
 	"go/token"
@@ -403,11 +404,14 @@ func linkMainPkg(ctx *context, pkg *packages.Package, pkgs []*aPackage, llFiles 
 		}
 	}()
 
-	// add rpath
+	// add rpath and find libs
 	exargs := make([]string, 0, ctx.nLibdir<<1)
+	libs := make([]string, 0, ctx.nLibdir*3)
 	for _, arg := range args {
 		if strings.HasPrefix(arg, "-L") {
 			exargs = append(exargs, "-rpath", arg[2:])
+		} else if strings.HasPrefix(arg, "-l") {
+			libs = append(libs, arg[2:])
 		}
 	}
 	args = append(args, exargs...)
@@ -418,6 +422,18 @@ func linkMainPkg(ctx *context, pkg *packages.Package, pkgs []*aPackage, llFiles 
 	}
 	err := ctx.env.Clang().Exec(args...)
 	check(err)
+
+	if runtime.GOOS == "darwin" {
+		dylibDeps := make([]string, 0, len(libs))
+		for _, lib := range libs {
+			dylibDep := findDylibDep(app, lib)
+			if dylibDep != "" {
+				dylibDeps = append(dylibDeps, dylibDep)
+			}
+		}
+		err := ctx.env.InstallNameTool().ChangeToRpath(app, dylibDeps...)
+		check(err)
+	}
 
 	switch mode {
 	case ModeRun:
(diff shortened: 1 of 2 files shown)


internal/build/build.go
```diff
@@ -419,6 +424,23 @@ func linkMainPkg(ctx *context, pkg *packages.Package, pkgs []*aPackage, llFiles
...
+		dylibPaths := make([]string, 0, len(libs))
+		rpaths := make([]string, 0, len(libs))
+		for _, lib := range libs {
+			dylibPath := findDylibPath(app, lib)
+			if dylibPath != "" {
+				dylibPaths = append(dylibPaths, dylibPath)
+				rpaths = append(rpaths, filepath.Dir(dylibPath))
+			}
+		}
+		installNameToolCmd := install_name_tool.New("")
```
xushiwei: use `ctx.env.InstallNameTool`

internal/build/build.go
```diff
@@ -419,6 +424,23 @@ func linkMainPkg(ctx *context, pkg *packages.Package, pkgs []*aPackage, llFiles
...
+		err := installNameToolCmd.ChangeToRpath(app, dylibPaths...)
+		check(err)
+		err = installNameToolCmd.AddRpath(app, rpaths...)
```
xushiwei: rpath is added already?

internal/build/build.go
```diff
@@ -403,11 +405,14 @@ func linkMainPkg(ctx *context, pkg *packages.Package, pkgs []*aPackage, llFiles
 		}
 	}()
 
-	// add rpath
+	// add rpath and find libs
 	exargs := make([]string, 0, ctx.nLibdir<<1)
+	libs := make([]string, 0, ctx.nLibdir*3)
```
xushiwei, line 409: Don't need save libs. Can use `debug/macho` to find all dylibs (see [dylibdeps](https://github.com/goplus/llgo/blob/main/chore/dylibdeps/dylibdeps.go)) and exclude those in `/lib/` or `/usr/lib`.

internal/build/build.go
aofei:
```
          cmd LC_RPATH
      cmdsize 56
         path /opt/homebrew/Cellar/bdw-gc/8.2.6/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path /opt/homebrew/opt/bdw-gc/lib (offset 12)
```

This step adds rpaths similar to `/opt/homebrew/opt/bdw-gc/lib` (the actual dylib ID), which is distinct from those derived from `-L` (e.g., the `/opt/homebrew/Cellar/bdw-gc/8.2.6/lib`).

internal/build/build.go
aofei: > exclude those in `/lib/` or `/usr/lib`.

My first thought was to apply rpath only to explicitly linked libs (those `-l` args). I'm not sure if using an exclude list is a good idea, considering not only `/lib` and `/usr/lib` need consideration, but also many other system-level libdirs like `/System/Library/Frameworks` and `/Library/Frameworks`, and possibly others.

internal/build/build.go
xushiwei: why we need so may rpath(s)?

internal/build/build.go
xushiwei: It's strange that we use prefix match to find real path of a dylib. If we want to do this, it seem using `os.Readlink` is better.

aofei:
```go
if dylib, ok := load.(*macho.Dylib); ok {
	if strings.HasPrefix(filepath.Base(dylib.Name), fmt.Sprintf("lib%s.", lib)) {
		return dylib.Name
	}
}
```

The `dylib.Name` here refers to the dylib ID.

Mach-O typically requires the dylib ID format to be `/path/to/lib<name>.<version>.dylib`. For our purposes, only the last path element needs to be matched, as it is unique (`clang` can ensures this for us). So this is why I do `filepath.Base(dylib.Name)`. And the prefix match is enough because the `<version>.dylib` part is not important is this case.

If we do fully match `os.Readlink(dylib.Name)`, then we may need to simulate `clang`'s process of searching `-L` to find `-l` to obtain the specific absolute path. I'm not sure if this is worth doing, since our use case here is pretty straight.

internal/build/build.go
aofei: This step is added because I believe it is more meaningful than the rpaths derived from `-L`, after all, they are dylib IDs and are usually not limited to specific versions.

But yes, that's too many rpaths (and will be more if dylibs are added). I'll remove this step.

PR #542, opened 2024-07-19
[wip] build: add cl.Context for linkname

visualfc:


cl/compile.go
```diff
@@ -79,22 +80,42 @@ type pkgInfo struct {
 
 type none = struct{}
 
+type Context struct {
+	prog     llssa.Program
+	link     sync.Map // pkgPath.nameInPkg => linkname
+	skipLine sync.Map // pkgPath => skip line
```
xushiwei: why we save `skipLine` not `skips`. there may have multiple `// llgo:skip` in a package.

visualfc: fixed, store pkg skipLines for support multi line.

PR #550, opened 2024-07-22
env:fix incorrect extra line breaks & subcmd match

luoliwoshang:
fix the issue of clang command errors caused by extra line breaks introduced by `exec.Command(cmd, args[1:]...).Output()`

----
This problematic command has an extra line break.
```
clang [-I/opt/homebrew/Cellar/llvm/18.1.8/include  -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
-emit-llvm -S -o /Users/zhangzhiyang/Library/Caches/go-build/56/5656941ebcccc22e088ca01e3b2d684aeba7737dc870cdab6d55dc63162bcb83-d.llcursor.cpp.ll -c /Users/zhangzhiyang/Documents/Code/goplus/llgo/c/clang/_wrap/cursor.cpp]
```
企业微信截图_11c77aba-3fd6-460a-b5d4-3df58c67f169

### DIFF
--- a/cl/_testlibc/demangle/in.go
+++ b/cl/_testlibc/demangle/in.go
@@ -1,0 +1,15 @@
+package main
+
+import (
+	"github.com/goplus/llgo/c"
+	"github.com/goplus/llgo/cpp/llvm"
+)
+
+func main() {
+	mangledName := "__ZNK9INIReader10ParseErrorEv"
+	if name := llvm.ItaniumDemangle(mangledName, true); name != nil {
+		c.Printf(c.Str("%s\n"), name)
+	} else {
+		println("Failed to demangle")
+	}
+}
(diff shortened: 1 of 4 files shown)


cl/_testlibc/llvm/in.go
```diff
@@ -0,0 +1,15 @@
+package main
```
xushiwei: don't call a main app `llvm`.

cl/internal/llvm/llvm.go
```diff
@@ -0,0 +1,16 @@
+package llvm
```
xushiwei: don't create this package. just use https://pkg.go.dev/github.com/goplus/llgo/cpp/llvm

PR #554, opened 2024-07-24
c/clang:symbol dump demo

luoliwoshang:
- [x] filename get null


c/clang/_wrap/cursor.cpp
```diff
@@ -1,29 +1,78 @@
...
 }
 
-extern "C" {
+extern "C"
+{
 
-CXString wrap_clang_getCursorSpelling(CXCursor *cur) {
-    return clang_getCursorSpelling(*cur);
-}
+    CXString wrap_clang_getCursorSpelling(CXCursor *cur)
```
xushiwei: use go style is better.

luoliwoshang: updated

c/clang/_wrap/cursor.cpp
```diff
@@ -1,29 +1,77 @@
...
 }
 
-extern "C" {
+extern "C"
+{
 
-CXString wrap_clang_getCursorSpelling(CXCursor *cur) {
-    return clang_getCursorSpelling(*cur);
-}
+    CXString wrapCursorSpelling(CXCursor *cur)
```
xushiwei: should have clang namespace. I don't think `wrapCursorSpelling` is better than `wrap_clang_getCursorSpelling`.

bug I think:

```
CXString wrap_clang_getCursorSpelling(CXCursor *cur) {
}
```

is better than:

```
    CXString wrap_clang_getCursorSpelling(CXCursor *cur)
    {
    }
```

Ident spaces is not need for a global function.

luoliwoshang: updated, I have configured clang-format as follows, which basically aligns with some Go styles:
```
    "C_Cpp.clang_format_style": "{ BasedOnStyle: LLVM, IndentWidth: 4, ColumnLimit: 120, BreakBeforeBraces: Custom, BraceWrapping: { AfterFunction: false } }"
```

PR #556, opened 2024-07-25
asyncio: implementation

cpunion:
- [x] POC
- [x] Non-blocking, remove chan and goroutine
  - [x] Parallel version
- [x] Scheduling last frame instead of scheduling ROOT task for performance
- [ ] Compiling async functions
  - [x] generator
  - [ ] promise
  - [ ] compitable recover block
- [ ] Support event IO


xushiwei:
Use `async` branch to polish it: https://github.com/goplus/llgo/tree/async

PR #572, opened 2024-07-26
fix(c/libuv): Fix return type FsType & struct rename

hackerchai:
Signed-off-by: hackerchai <i@hackerchai.com>

### DIFF
--- a/c/libuv/libuv.go
+++ b/c/libuv/libuv.go
@@ -297,7 +297,7 @@ func (handle *Handle) Fileno(fd *OsFd) c.Int {
 }
 
 //go:linkname UvPipe C.uv_pipe
-func UvPipe(fds [2]UvFile, readFlags c.Int, writeFlags c.Int) c.Int {
+func UvPipe(fds [2]File, readFlags c.Int, writeFlags c.Int) c.Int {
 	return 0
 }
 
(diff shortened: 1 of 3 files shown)


c/libuv/libuv.go
```diff
@@ -297,7 +297,7 @@ func (handle *Handle) Fileno(fd *OsFd) c.Int {
 }
 
 //go:linkname UvPipe C.uv_pipe
-func UvPipe(fds [2]UvFile, readFlags c.Int, writeFlags c.Int) c.Int {
+func UvPipe(fds [2]File, readFlags c.Int, writeFlags c.Int) c.Int {
```
aofei:
```suggestion
func Pipe(fds [2]File, readFlags c.Int, writeFlags c.Int) c.Int {
```

hackerchai: There is a repeated Struct named `pipe`, so we rename the func into `UvPipe`
- `libuv.go`
```go
type Pipe struct {
	Unused [0]byte
}
```

xushiwei: maybe: type `uv_pipe_t` => `PipeT`, func  `uv_pipe` => `Pipe`.

hackerchai: > maybe: type `uv_pipe_t` => `PipeT`, func `uv_pipe` => `Pipe`.

solved due to srtuct `Pipe`removed.

Issue #619, opened 2024-07-30
Unexpected Panic Execution Causing Compile Error


luoliwoshang:
This bug was first discovered in the following Pull Request:
[https://github.com/goplus/llgo/pull/566](https://github.com/goplus/llgo/pull/566)

The code provided below is a minimal reproduction of this issue.
```go
package main

import "errors"

func main() {
	syms := []int{}
	m := make(map[int]bool)
        // m[0] = true // also with error

	// Uncommenting the following will prevent llgo run . from crashing
	// for i := range syms {
	// 	_ = i
	// }

	err := errors.New("expect error")

	// Normal execution
	// if err != nil {
	// 	panic(err)
	// }

	// Erroneous execution (uncomment to simulate error)
	check(err)

	// comment to prevent error
	for range syms {
		_, ok := m[0]

		// comment to prevent error
		if ok {
			println("ok")
		}
	}

	// comment to prevent error
	defer println("bye")

	// comment to prevent error
	for _, sym := range syms {
		_ = sym
	}
}

func check(err error) {
	if err != nil {
		panic(err)
	}
}
```

### Expected Behavior
The program is expected to output `panic: failed to read symbol table`, consistent with a direct panic in the function.
### Actual Behavior 
The program's behavior exhibits two distinct patterns:
**Crash Scenario:**

When the code is as shown in the minimal reproduction code and `check(err)` is called, `llgo install .` `llgo run .` 
crashed

**Non-Crash Scenario:**
The following modification resolves the crash issue when check(err) is called:
a. Uncommenting the first for loop:
```go
for i := range syms {
      _ = i
}
```
However, making any of the following modifications causes the program to crash when `check(err)` is called:
b. Commenting out the second for loop:
```go
// for range syms {
// 	_, ok := m[0]
// 	if ok {
// 		println("ok")
// 	}
// }
```
c. Commenting the `if` in the  second loop
The crash is also related to the if branch following the access of a map key within a for range loop.
```go
for range syms {
    _, ok := m[0]
    // comment to prevent error
    // if ok {
    // 	println("ok")
    // }
}
```
d. Commenting out the defer statement:
```go
// defer println("bye")
```
e. Commenting out the last for loop:
```go
// for _, sym := range syms {
//     _ = sym
// }
```
### Error Output
When the crash occurs, the following error is output:
```bash
_xtool/llcppsymg/bugtest on  llcppg/symg/header [$!?] via 🐹 v1.20.4 
❯ llgo run .
[...]

PR #643, opened 2024-08-02
 ssa: fix defer move block[0] index

visualfc:
fix https://github.com/goplus/llgo/issues/619


luoliwoshang:
> fix #619

run normally

PR #656, opened 2024-08-05
library: net/url

spongehah:


_cmptest/urldemo/url.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/url"
+)
+
+func main() {
+	u, err := url.Parse("http://bing.com/search?q=dotnet")
```
aofei, line 10:
```suggestion
	u, err := url.Parse("https://example.com/foo?bar=1")
```

Do not use such commercial services. Use `example.com` in tests or demos.

aofei (approved):

PR #700, opened 2024-08-12
refactor(c/net): Use c link instead of manual implement func

hackerchai:


hackerchai:
@xushiwei ready for preview.

PR #706, opened 2024-08-13
fix(c/libuv): Add multiple struct size

hackerchai:


hackerchai:
@xushiwei ready for preview.

PR #707, opened 2024-08-14
lib/net/textproto: patch Dial

spongehah:


internal/build/_overlay/net/textproto/textproto.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	net.Freeaddrinfo(res)
+	return &cConn{
+		socketFd: socketFd,
+	}, nil
+}
+
+func splitAddr(addr string) (host, port string, err error) {
+	// 处理 IPv6 地址
```
aofei: Comments should be in English only.

internal/build/_overlay/net/textproto/textproto.go
```diff
@@ -0,0 +1,254 @@
...
+func (c *cConn) Write(p []byte) (n int, err error) {
+	result := os.Write(c.socketFd, unsafe.Pointer(&p[0]), uintptr(len(p)))
+	if result < 0 {
+		return 0, fmt.Errorf("write error")
+	}
+	return result, nil
+}
+
+func (c *cConn) Close() error {
+	c = nil
```
aofei: Why don't we use `os.Close(c.socketFd)`?

And, what happens if the caller attempts to `Read` or `Write` after `Close`?

internal/build/_overlay/net/textproto/textproto.go
```diff
@@ -0,0 +1,254 @@
...
+		return nil, err
+	}
+	var hints net.AddrInfo
+	var res *net.AddrInfo
+	c.Memset(unsafe.Pointer(&hints), 0, unsafe.Sizeof(hints))
+	hints.Family = net.AF_UNSPEC
+	hints.SockType = net.SOCK_STREAM
+	status := net.Getaddrinfo(c.AllocaCStr(host), c.AllocaCStr(port), &hints, &res)
+	if status != 0 {
+		return nil, fmt.Errorf("getaddrinfo error")
```
aofei:
```suggestion
		return nil, errors.New("getaddrinfo error")
```

internal/build/_overlay/net/textproto/textproto.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return result, nil
+}
+
+func (c *cConn) Write(p []byte) (n int, err error) {
+	result := os.Write(c.socketFd, unsafe.Pointer(&p[0]), uintptr(len(p)))
+	if result < 0 {
+		return 0, fmt.Errorf("write error")
+	}
+	return result, nil
```
aofei: Refer to https://pkg.go.dev/io#Writer:

> Write must return a non-nil error if it returns n < len(p).

internal/build/_overlay/net/textproto/textproto.go
```diff
@@ -0,0 +1,254 @@
...
+	return host, port, nil
+}
+
+func (c *cConn) Read(p []byte) (n int, err error) {
+	result := os.Read(c.socketFd, unsafe.Pointer(&p[0]), uintptr(len(p)))
+	if result < 0 {
+		return 0, fmt.Errorf("read error")
+	}
+	if result == 0 {
+		return 0, io.EOF
```
aofei: What happens if `os.Read` reads `0 < result < len(p)` and then reaches the end (aka an early EOF)?

internal/build/_overlay/net/textproto/textproto.go
```diff
@@ -0,0 +1,254 @@
...
+		port = "80" // Default port is 80
+	}
+
+	return host, port, nil
+}
+
+func (c *cConn) Read(p []byte) (n int, err error) {
+	result := os.Read(c.socketFd, unsafe.Pointer(&p[0]), uintptr(len(p)))
+	if result < 0 {
+		return 0, fmt.Errorf("read error")
```
aofei: Shouldn't we utilize `syscall.Errno` here?

internal/build/_overlay/net/textproto/textproto.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+type cConn struct {
+	socketFd c.Int
+	closed   bool
+}
+
+func (conn *cConn) Read(p []byte) (n int, err error) {
+	if conn == nil || conn.closed {
+		return 0, errors.New("connection is closed")
```
aofei:
```suggestion
		return 0, fs.ErrClosed
```

https://pkg.go.dev/io/fs#ErrClosed

aofei (changes requested):

PR #731, opened 2024-08-20
llcppg/ast:TagExpr

luoliwoshang:


chore/llcppg/ast/ast.go
```diff
@@ -285,7 +285,7 @@ func (*FuncDecl) declNode() {}
 // struct/union/class Name { Field1, Field2, ... };
 type TypeDecl struct {
 	DeclBase
-	Tag     Tag
+	Tag     TagExpr
```
xushiwei: `TypeDecl` is not a `TagExpr`. add a `Name *Ident` directly.

luoliwoshang: > `TypeDecl` is not a `TagExpr`. add a `Name *Ident` directly.

When should use a  `TagExpr`

PR #751, opened 2024-08-28
fix(c/libuv): Fix EOF error value

hackerchai:


hackerchai:
@xushiwei ready for review

PR #767, opened 2024-09-04
Improve the llgo cjson library

tsingbx:


c/cjson/cjson.go
```diff
@@ -26,128 +26,650 @@ const (
 	LLGoPackage = "link: $(pkg-config --libs libcjson); -lcjson"
 )
 
+type JSON_bool c.Int
```
xushiwei: use `Bool` instead of `JSON_bool`

c/cjson/cjson.go
```diff
@@ -26,128 +26,650 @@ const (
...
+
 // llgo:type C
 type JSON struct {
 	Unused [0]byte
 }
 
+func ParseBytes(value []byte) *JSON {
+	return ParseWithLength((*c.Char)(unsafe.Pointer(&value[0])), uintptr(len(value)))
```
xushiwei: use `unsafe.SliceData` instead of `&value[0]`

c/cjson/cjson.go
```diff
@@ -26,128 +26,650 @@ const (
...
+}
+
+func ParseString(value string) *JSON {
+	return ParseWithLength(c.AllocaCStr(value), uintptr(len(value)))
```
xushiwei: use `unsafe.StringData` instead of `AllocaCStr`.

c/cjson/cjson.go
luoliwoshang: > use `Bool` instead of `JSON_bool`

Will conflict with this function name
```go
//go:linkname Bool C.cJSON_CreateBool
func Bool(boolean c.Int) *JSON
```

c/cjson/cjson.go
```diff
@@ -450,57 +445,57 @@ func CreateDoubleArray(numbers *c.Double, count c.Int) *JSON
 //go:linkname CreateStringArray C.cJSON_CreateStringArray
 func CreateStringArray(strings *c.Char, count c.Int) *JSON
 
-// CJSON_PUBLIC(cJSON_bool) cJSON_AddItemToArray(cJSON *array, cJSON *item);
+// CJSON_PUBLIC(cBool) cJSON_AddItemToArray(cJSON *array, cJSON *item);
```
luoliwoshang: If the original API declaration is retained, then it should still be cJSON_bool here.

c/cjson/cjson.go
```diff
@@ -26,128 +26,645 @@ const (
...
+// llgo:link (*JSON).DeleteItemFromObjectCaseSensitive C.cJSON_DeleteItemFromObjectCaseSensitive
+func (o *JSON) DeleteItemFromObjectCaseSensitive(s *c.Char) {}
+
+// CJSON_PUBLIC(cJSON_Bool) cJSON_InsertItemInArray(cJSON *array, int which, cJSON *newitem);
+//
+// Update array items.
+// Shifts pre-existing items to the right.
+//
+// llgo:link (*JSON).InsertItemInArray C.cJSON_InsertItemInArray
+func (o *JSON) InsertItemInArray(which c.Int, newitem *JSON) Bool { return Bool(0) }
```
luoliwoshang, line 543: it is better to remove the original function signature once the mapping is complete, just keep the original comments

PR #776, opened 2024-09-06
llcppg/ast:elaborate type's uid

luoliwoshang:
When a typedef assigns multiple aliases to the same anonymous RecordType or EnumType, it creates a corresponding number of embedded RecordTypes or EnumTypes. However, these actually point to the same underlying type. A uid (Unique Identifier) attribute is added to these ElaboratedTypes, expected to be represented using the Unified Symbol Resolution (USR) format.
#### Example
```c
typedef struct { int x; int y; } Point, *PointPtr;
```
#### Related Api
```c
/**
 * Retrieve a Unified Symbol Resolution (USR) for the entity referenced
 * by the given cursor.
 *
 * A Unified Symbol Resolution (USR) is a string that identifies a particular
 * entity (function, class, variable, etc.) within a program. USRs can be
 * compared across translation units to determine, e.g., when references in
 * one translation refer to an entity defined in another translation unit.
 */
CINDEX_LINKAGE CXString clang_getCursorUSR(CXCursor);
```

### DIFF
--- a/chore/llcppg/ast/ast.go
+++ b/chore/llcppg/ast/ast.go
@@ -247,6 +247,7 @@ func (*FuncType) exprNode() {}
 // ------------------------------------------------
 
 type RecordType struct {
+	Uid     string // Unified Symbol Resolution (USR)
 	Tag     Tag
 	Fields  *FieldList
 	Methods []*FuncDecl
@@ -298,6 +299,7 @@ type EnumItem struct {
 func (*EnumItem) exprNode() {}
 
 type EnumType struct {
+	Uid   string // Unified Symbol Resolution (USR)
 	Items []*EnumItem
 }
 


chore/llcppg/ast/ast.go
```diff
@@ -247,6 +247,7 @@ func (*FuncType) exprNode() {}
 // ------------------------------------------------
 
 type RecordType struct {
+	Uid     string // Unified Symbol Resolution (USR)
```
xushiwei, line 250: 在哪里使用它？需要有 ast 的节点引用它

luoliwoshang: RecordDecl会存在RecordType，并且这个RecordType实际上在libclang实际上都会存在一个节点（这个类型是具名的，那么这个type的usr就引用自具名的RecordDecl，如果是匿名的类型，那么就会引用一个匿名的RecordDecl）

也就是说Recordtype类型一定会有一个对应的ast节点在libclang中

PR #781, opened 2024-09-09
Future IO update

cpunion:
* Design doc update
* Future.Then
* Support multi callbacks/awaiters


x/async/README.md
```diff
@@ -362,28 +362,66 @@ In some situations, you may want to get the first result of multiple async opera
...
-- For performance reason don't implement async functions with goroutines
+- For performance and memory reasons don't implement async functions with goroutines
 - Avoid implementing `Promise` by using `chan` to avoid blocking the thread, but it can be wrapped as a `chan` to make it compatible `select` statement
 
 ## Design
 
-Introduce `async.IO[T]` type to represent an asynchronous operation, `async.Future[T]` type to represent the result of an asynchronous operation. `async.IO[T]` can be `bind` to a function that accepts `T` as an argument to chain multiple asynchronous operations. `async.IO[T]` can be `await` to get the value of the asynchronous operation.
+### `async.Future[T]` type
+
+Introduce `async.Future[T]` type to represent an asynchronous operation that will produce a value of type `T`, similar to `Promise`/`Future` in other languages.
```
nighca: As far as I know, `Promise` in Javascript represents the "(coming) result" of the asynchronous operation, instead of the asynchronous operation itself.

The difference makes `async.Future` here very different with `Promise` in Javascript. It matters for the decision of "when should the asynchronous operation start". In Javascript, typically the asynchronous operation starts when the promise is created, or before it is created.

While in design here, the `async.Future` represents the operation itself, which means the asynchronous operation does not start when the `async.Future` object is created. It starts when its `Then` is called or some one calls `async.Run` with it. That makes it more like a "promise factory", which I think may be unexpected.

cpunion: The executor usage (`async.Run` here) will be changed, maybe hides `async.Run` and provides global hidden executor to auto start asynchronous calls. And the `Then` driven execution will be removed.

PR #788, opened 2024-09-10
llcppsymg:refine class fetch,test case,-v

luoliwoshang:
[wip] fix https://github.com/goplus/llgo/issues/830 

* Refactored the class name retrieval logic, now  extracting it by accessing the cursor's semantic parent, and removed redundant logic
* supplement test cases
* Due to the different number of leading underscores in C++ symbols on Linux and macOS, previously removing only the first leading underscore would cause comparison failures on Linux. Therefore, to accommodate both scenarios, we remove remove first leading underscore only in mac
* For paths without -L,llcppsymg attempts to match and retrieve from the default search paths &  .so in linux,.dylib in macos
* -v flag
* find dylib&headerfile from mutiple path

### linux 
- [ ] wait os.Readdir  to search `/etc/ld.so.conf.d` 's subdir
- [ ] wait exec.Command not panic

It can pass tests normally in both Linux and macOS environments.
However, in actual operation on Linux, the part calling nm throws the following error. Investigation reveals that it's caused by exec.Command. Running `cmptest/osexec` on Linux also produces the corresponding issue:

```bash
root@be00d9b1c2c9:~/llgo/chore/_xtool/llcppsymg# llgo run .
parse dylib symbols from config lib:-L/lib/aarch64-linux-gnu -llua5.4
panic: todo: syscall.Faccessat
```


xushiwei:
should resolve conflicts.

luoliwoshang:
> should resolve conflicts.

resolve conflicts , and need merge the https://github.com/goplus/llgo/pull/757 for llcppsigfetch & gogensig

luoliwoshang:
ready for review. @xushiwei

PR #799, opened 2024-09-19
fix(c/libuv): Add missing essential funcs for uv_loop_t & uv_req_t

hackerchai:


hackerchai:
@xushiwei ready for review

PR #812, opened 2024-09-29
llcppsigfetch:cross-platform,log,fix bugs

luoliwoshang:
* llcppg:without this close, cause the program to hang.
* The default signedness of `char` can vary across different platforms. Therefore, test for Char confirms whether the type is `char` and verifies if its flag is either signed or unsigned. Additionally, to ensure that comparison tests for structures can pass on Linux as well,removed the use cases where char types are referenced within structures.
* refine args parse
* -v
### Bugs
- [x] Fix a special case where libclang parses FILE type reference in fopen function, resolving its null return value    https://github.com/luoliwoshang/llgo/issues/106
- [x] Fix anonymous enumdecl display error https://github.com/luoliwoshang/llgo/issues/110


luoliwoshang:
ready for review. @xushiwei

PR #835, opened 2024-10-19
modify llcppg command to run go mod init and go get

tsingbx:
If you do not add it, you cannot run the gogensig command successfully.


tsingbx:
movo to gogensig

PR #863, opened 2024-11-12
fix llcppsigfetch: invalid memory address or nil pointer dereference

tsingbx:


tsingbx:
Fixed crash when writing to standard output when the converted C header file contains comments with format strings like %s, %d, etc.

PR #893, opened 2024-11-29
add clang.GoString, clang.File functions for helper of llcppsymg struct_methodname

tsingbx:


c/clang/basic.go
```diff
@@ -57,3 +57,11 @@ type StringSet struct {
  */
 // llgo:link (*StringSet).Dispose C.clang_disposeStringSet
 func (*StringSet) Dispose() {}
+
+func GoString(clangStr String) (str string) {
+	defer clangStr.Dispose()
+	if clangStr.CStr() != nil {
```
xushiwei: use a temp string to reduce function call.

PR #923, opened 2024-12-27
[WIP] ssa: compatable with C function calls that have struct parameters.

visualfc:
compatible c abi struct calling.

-  link c abi wrap ( c wrap code generate )
-  callback abi wrap ( go x/ffi wrap func by libffi )
-  TODO cgo mode abi wrap

```
typedef struct {
    int a[2];
    int b;
} info;

typedef struct {
    int a[128];
} info_big;

info demo1(info i, int n);
info_big big1(info_big b, int n);

define { i64, i32 } @demo1(i64 %0, i32 %1, i32 noundef %2) #0 
define void @big1(ptr dead_on_unwind noalias writable sret(%struct.info_big) align 4 %0, ptr noundef byval(%struct.info_big) align 8 %1, i32 noundef %2) #0 
```
generate wrap c code
```
typedef struct {GoInt32 f0[2];GoInt32 f1;} info;
extern info demo1(info p0, GoInt32 p1);
void llgo_wrapabi_demo1(info *p0, GoInt32 p1, info *r) {
    *r = demo1(*p0, p1);
}
typedef struct {GoInt32 f0[128];} bigArr;
extern bigArr big1(bigArr p0, GoInt32 p1);
void llgo_wrapabi_big1(bigArr *p0, GoInt32 p1, bigArr *r) {
    *r = big1(*p0, p1);
}
```

**TODO** check c/clang ( remove wrap )  https://github.com/visualfc/llgo/tree/wrap_clang
    - pass on macOS amd64
    - _**panic on macOS arm64**_


cpunion:
Why not change calling convention to cdecl? This way will add function calling.

Seems there is an API to change CC, but I'm not sure if it change the params passing, I think it should change it:
```c
Function *F = ...;
F->setCallingConv(llvm::CallingConv::C);
```

The other way is processing args before calling and in function, may be more complex, but seems this PR is complex too.

cpunion:
> Why not change calling convention to cdecl? The way in this PR will add function calling.
> 
> Seems there is an API to change CC, but I'm not sure if it changes the params passing, I think it should change it:
> 
> ```c
> Function *F = ...;  
> F->setCallingConv(llvm::CallingConv::C);
> ```
> 
> The other way is processing args before calling and in function, may be more complex, but I guess it's simpler than this PR.

I means change all funcs to CC, but maybe the easy way is just change de C func calling, don't need warp it.

cpunion:
> I mean change all funcs to CC, but maybe the easier way is just change the C func calling (no need process params in function), and don't need warp it.

Maybe LLVM doesn't process struct params passing, I saw clang outputed LLVM-IR copies structs on stack and passes address to the callee function, if the struct size is larger than 2 ptr size.

PR #937, opened 2025-01-06
README: fix dependences requirement for macOS

MeteorsLiu:
User may not install pkg-config when initalization.

When lacking of pkg-config tool, it will panic like that 

```
panic: test link args ' $(pkg-config --libs bdw-gc); -lgc' failed
	expanded to: [-lgc]
	resolved to: [-lgc]
	error: ld: library 'gc' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,7 +359,7 @@ Follow these steps to generate the `llgo` command (its usage is the same as the 
 
 ```sh
 brew update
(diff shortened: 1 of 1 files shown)


cpunion:
doc/_readme/scripts/install_macos.sh need update

PR #959, opened 2025-01-14
syscall(linux): forkAndExecInChild

luoliwoshang:
#### result
before use exec.command in linux
**got**
```
panic("todo: syscall.forkAndExecInChild")
```
**now**
```
root@be00d9b1c2c9:~/llgo/_demo/commandrun# llgo run .
len: 62
data: total 4
-rw-r--r-- 1 root root 258 Jan 16 10:52 commandrun.go
```
in the ci , use the exec.command to call nm tool,it can work normally.https://github.com/goplus/llcppg/pull/160

#### resolution
- The libc wrapper's clone function cannot create processes with an empty child stack. Therefore, we need to directly use syscall to invoke the system's clone function instead of using libc's clone implementation.

- On Linux x86, when creating a process using clone with CLONE_VM flag (similar to Go's approach), it sets up memory sharing between parent and child processes. This can lead to unexpected parent process termination. Go addresses this on Linux x86 through assembly optimizations, allowing parent and child processes to share memory while ensuring that executing exec doesn't clear the shared stack, thus enabling normal execution.

#### question
- [x] in ci, the demo will fail,but in my local docker env , can run success https://github.com/goplus/llgo/actions/runs/12826009988/job/35765102523?pr=959
#### other question(resolved)
in linux, go use clone to fork a child process,current is temp use fork to create a child process
`src/syscall/exec_linux.go`
```go
	locked = true
	if clone3 != nil {
		pid, err1 = rawVforkSyscall(_SYS_clone3, uintptr(unsafe.Pointer(clone3)), unsafe.Sizeof(*clone3))
	} else {
		flags |= uintptr(SIGCHLD)
		if runtime.GOARCH == "s390x" {
			// On Linux/s390, the first two arguments of clone(2) are swapped.
			pid, err1 = rawVforkSyscall(SYS_CLONE, 0, flags)
		} else {
			pid, err1 = rawVforkSyscall(SYS_CLONE, flags, 0)
		}
	}
```
in above code,only pass two param to clone, but in the signature of c,have three params
```c
int clone (int (*__fn) (void *__arg), void *__child_stack, int __flags, void *__arg, ...) ;
```
and if we only pass the flags & arg through call c.clone
```go
err1 = Errno(os.Clone(nil, 0, c.Int(flags), 0))
fmt.Println(err1)
fmt.Println(Errno(os.Errno()))
```
```go
// llgo:type C
type fn func(__arg c.Pointer) c.Int

//go:linkname Clone C.clone
func Clone(__fn fn, __childStack c.Pointer, __flags c.Int, __arg uintptr, __llgo_va_list ...any) c.Int
```
will got 
```
Unknown error -1
Invalid argument
```
its error output like use clone without child_stack , got Invalid argument
```
int main() {
    const int STACK_SIZE = 1024 * 1024;
    void *stack = malloc(STACK_SIZE);
    if (!stack) {
        perror("malloc");
        return 1;
    }

    int flags = SIGCHLD;
    // int pid = clone(child_function, stack + STACK_SIZE, flags, NULL);  work normally
    int pid = clone(child_function, NULL, flags, NULL); // got clone: Invalid argument
    if (pid == -1) {
        perror("clone");
        free(stack);
        return 1;
    }

    printf("Parent process PID: %d, Child PID: %d\n", getpid(), pid);
    free(stack);
    return 0;
}
```
in the under asm code，it have some operate
```asm
// func rawVforkSyscall(trap, a1, a2 uintptr) (r1, err uintptr)
TEXT ·rawVforkSyscall(SB),NOSPLIT|NOFRAME,$0-40
	MOVQ	a1+8(FP), DI
	MOVQ	a2+16(FP), SI
	MOVQ	$0, DX
	MOVQ	$0, R10
	MOVQ	$0, R8
	MOVQ	$0, R9
	MOVQ	trap+0(FP), AX	// syscall entry
	POPQ	R12 // preserve return address
	SYSCALL
	PUSHQ	R12
	CMPQ	AX, $0xfffffffffffff001
	JLS	ok2
	MOVQ	$-1, r1+24(FP)
	NEGQ	AX
	MOVQ	AX, err+32(FP)
	RET
ok2:
	MOVQ	AX, r1+24(FP)
	MOVQ	$0, err+32(FP)
```
so in current resolution, use the syscall to call the clone instead the libc wrapper clone;
 from https://github.com/goplus/llgo/pull/959#issuecomment-2597522308 's resolution
```
//go:linkname Syscall C.syscall
func Syscall(sysno c.Long, __llgo_va_list ...any) c.Long
```
```
ret := os.Syscall(syscall.SYS_CLONE, flags, 0)
```


MeteorsLiu:
The wrapper is accurate, the reason why the function is different from Go's is that libc wraps Clone in its way.

Raw Syscall may like(x86-64):
```c
 long clone(unsigned long flags, void *stack,
                      int *parent_tid, int *child_tid,
                      unsigned long tls);
```

In libc, it may like:

```c
 int clone(int (*fn)(void *_Nullable), void *stack, int flags,
                 void *_Nullable arg, ...  /* pid_t *_Nullable parent_tid,
                                              void *_Nullable tls,
                                              pid_t *_Nullable child_tid */ );
```

Acutally, the stack can be nullable, however, libc does a force verification for stack:

Code from gblic

```
[ENTRY](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/ENTRY) ([__clone](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/__clone))
	[cmp](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/cmp)	[r0](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/r0), 0		/* @fn can't be NULL.  */
	and	[r1](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/r1),[r1](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/r1),-4	/* @child_stack be 4 bytes aligned per ABI.  */
	[cmp](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/cmp).ne	[r1](https://elixir.bootlin.com/glibc/glibc-2.40.9000/C/ident/r1), 0		/* @child_stack can't be NULL.  */
	bz	L (__sys_err)
```

In libc, stack cannot be nullable.

luoliwoshang:
> In libc, stack cannot be nullable because of calling the fn.
> 
> When you don't need to call fn, the stack param can be zero.

Good investgation! We may use the c.syscall to call the clone function from core like go vForkSyscall , to avoid libc's wrapper  check.

luoliwoshang:
when excute clone with log output
```go
println("Attempting clone with flags:", flags)
			ret := os.Syscall(syscall.SYS_CLONE, flags, 0)
			if ret >= 0 {
				if ret == 0 {
					println("clone succeeded, now in child")
				} else {
					println("clone succeeded now in parent")
				}
				pid = uintptr(ret)
				err1 = Errno(0)
			} else {
				pid = 0
				err1 = Errno(os.Errno())
				println("clone failed", err1)
			}
```
in my docker's linux will got
```
VSCODE_IPC_HOOK_CLI=/tmp/vscode-ipc-07733c65-1a88-47ec-87d2-10a6579cd33d.sock
_=/root/go/bin/llgo
Attempting clone with flags: 16657
clone succeeded, now in child
Pass 1: look for fd[i] < i and move those up above len(fd)
Pass 2: dup fd[i] down onto i.
clone succeeded now in parent
len: 11
data: hello llgo
```
but in ci ,only got follow
https://github.com/goplus/llgo/actions/runs/12859660677/job/35850417459?pr=959
```
_=/home/runner/go/bin/llgo
OLDPWD=/home/runner/work/llgo/llgo
Attempting clone with flags: 16657
clone succeeded, now in child
Pass 1: look for fd[i] < i and move those up above len(fd)
Pass 2: dup fd[i] down onto i.
os.Execve(argv0, argv, envv) 2
```

luoliwoshang:
when set the clone flag only use `CLONE_VFORK` , it can run normally 

https://github.com/goplus/llgo/actions/runs/12862918395/job/35858617503?pr=959
```
// The Go origin implement is:
// flags |= CLONE_VFORK | CLONE_VM
// llgo use CLONE_VFORK only.
// https://www.man7.org/linux/man-pages/man2/clone.2.html
flags |= CLONE_VFORK
```

PR #963, opened 2025-01-18
Find llgo root

cpunion:
- [x] Find llgo root for easy development (don't need set LLGO_ROOT)


cpunion:
Wait https://github.com/goplus/llgo/pull/954

PR #1013, opened 2025-02-26
add all llcppg.pub for clib

tsingbx:


c/os/llcppg.pub
```diff
@@ -0,0 +1,7 @@
+mode_t ModeT
+uid_t UidT
+gid_t GidT
+off_t OffT
+dev_t DevT
+Stat_t StatT
```
luoliwoshang, line 6: c origin name is `stat`~

c/pthread/llcppg.pub
```diff
@@ -0,0 +1,3 @@
+pthread_t Thread
+attr Attr
+pthread_key_t Key
```
xushiwei, line 3: KeyT

PR #1038, opened 2025-03-24
feat: add `llgo get`

seri037:
- [x] add `mod` package to sync and query `llpkgstore.json`
- [x] complete logic of `llgo get`
- [ ] setup proper `PKG_CONFIG_PATH` when `llgo build`


compiler/internal/mod/metadata.go
```diff
@@ -0,0 +1,148 @@
...
+
+	err = m.buildVersionsHash()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *metadataMgr) buildVersionsHash() error {
```
luoliwoshang: The function name `buildVersionsHash` might be misleading. What it actually builds is just a mapping. Using "Hash" to describe this mapping functionality could be misinterpreted as performing hash encryption on this JSON.

compiler/internal/mod/mod.go
```diff
@@ -0,0 +1,126 @@
...
+
+func doPathCompletion(name, goVer string) (string, error) {
+	major := semver.Major(goVer)
+
+	if major == "" {
+		return "", fmt.Errorf("not a semver: %s", goVer)
+	} else if major == "v0" || major == "v1" {
+		return fmt.Sprintf("%s/%s", LLPkgPathPrefix, name), nil
+	} else {
+		return fmt.Sprintf("%s/%s/v%s", LLPkgPathPrefix, name, major), nil
```
luoliwoshang: duplicates the "v" prefix. Since semver.Major already returns values with "v" (e.g., "v2"), this creates paths with double "v" (like "/vv2").

compiler/internal/get/get.go
```diff
@@ -0,0 +1,116 @@
...
+	isLLPkg, err := mod.IsLLPkg(actualModule)
+	if err != nil {
+		return err
+	} else if !isLLPkg {
+		// no more work
+		return nil
+	}
+
+	// 3. Now we need to parse the llpkg.cfg file.
+	cfgPath, err := mod.LLPkgCfgFilePath(actualModule)
```
luoliwoshang: Suggest optimizing the usage of `mod.IsLLPkg` in Do method. Currently it calls `mod.IsLLPkg` to check if it's an LLPkg, then calls LLPkgCfgFilePath to get the path. Since `mod.IsLLPkg` internally calculates the config file path as well, suggest modifying `mod.IsLLPkg` to:
```go
func IsLLPkg(module string) (isLLPkg bool, cfgPath string, err error)
```
This would avoid calculating the config file path twice in the Do method.

compiler/internal/mod/mod.go
seri037: donne at [fix: duplicated 'v' prefix of semver](https://github.com/goplus/llgo/pull/1038/commits/38d8f893dad3b4df8b6d985184a2e2e453d51458)

compiler/internal/mod/metadata.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/720cc2ec217510c3bde044f2023591ed6bcef012

seri037:
need review @visualfc @aofei

compiler/internal/get/get.go
seri037: Using multiple return values can lead to misunderstandings. Instead, a direct approach is preferable since the  performance difference can be insignificant.

compiler/go.mod
```diff
@@ -6,16 +6,15 @@ require (
 	github.com/goplus/gogen v1.16.9
 	github.com/goplus/llgo v0.9.9
 	github.com/goplus/llgo/runtime v0.0.0-00010101000000-000000000000
+	github.com/goplus/llpkgstore v0.0.0-20250318065259-1f6d7d1376ae
```
xushiwei: don't depend this repo.

seri037: We integrate some llpkg-related functionalities (e.g. version query, binary download) into `goplus/llpkgstore` because they're going to be used by other components, that is `pkgsite`, `GitHub Actions`. It can also reduce the redundancy in code.

xtool/env/env.go
```diff
@@ -61,6 +61,7 @@ func expandEnvWithCmd(s string) (string, bool) {
 
 		var out []byte
 		var err error
+
```
luoliwoshang: unrelated line update

seri037: done at https://github.com/goplus/llgo/pull/1038/commits/5729a2607fb3b9460ceac379264295048225d09b

compiler/go.mod
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/c06603154fc029f04a313517b7924d3012c52669

compiler/internal/mod/cache.go
```diff
@@ -0,0 +1,166 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+// Cache represents a local cache for storing and retrieving data.
+// It binds a local file path and a remote data source URL
+type cache[T any] struct {
+	data T // stores cached data
```
luoliwoshang, line 17: Given that the `cache` currently has only one user, perhaps we could simplify the implementation by temporarily removing the generic design?

seri037: sure `cache` has only one user, but using generic design keeps extensibility. I don't think it a proper way to integrate type of `metadata` directly into `cache`

compiler/cmd/internal/get/get.go
```diff
@@ -18,12 +18,18 @@
...
+
 	"github.com/goplus/llgo/compiler/cmd/internal/base"
+	"github.com/goplus/llgo/compiler/internal/get"
+	"github.com/goplus/llgo/compiler/internal/mockable"
 )
 
 // llgo get
 var Cmd = &base.Command{
-	UsageLine: "llgo get [-t -u -v] [build flags] [packages]",
+	UsageLine: "llgo get [clibs/packages]",
```
visualfc: llgo get 原有的 flags 可以保留支持，传递到 go get 中。

compiler/internal/env/env.go
```diff
@@ -118,3 +118,21 @@ func getRuntimePkgDirByCaller() (string, error) {
...
+
+	// If not set, LLGOCACHE defaults to {UserCacheDir}/llgo/
+	userCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		panic("Can not get user cache dir: " + err.Error())
+	}
+	return filepath.Join(userCacheDir, "llgo")
+}
+
+func LLPkgCache() string {
```
visualfc: 没有使用到的函数在这个 PR 本身不需要定义

compiler/cmd/internal/get/get.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/0041c777c2c5d977a5eab5e82caabc4bbab489fe

compiler/internal/env/env.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/b56a73a6a369461b7ecf448f0df4696983159e3d

compiler/internal/get/get.go
aofei, line 1: Consider moving from `compiler/internal/get` to `compiler/internal/modget` for better semantics.

compiler/cmd/internal/get/get.go
```diff
@@ -32,5 +38,52 @@ func init() {
...
+	modules := []string{}
+	if flagEndIndex >= 0 {
+		modules = args[flagEndIndex+1:]
+	} else {
+		modules = args
+	}
+
+	for _, m := range modules {
+		// Extract the name/path and version from the first argument
+		name, version := parse(m)
```
aofei:
```suggestion
		name, version, _ := strings.Cut(m, "@")
```

And remove `parse` as it is no longer needed.

compiler/internal/env/env.go
```diff
@@ -118,3 +118,17 @@ func getRuntimePkgDirByCaller() (string, error) {
...
+func LLGOCACHE() string {
+	// Check if LLGOCACHE is set
+	if llgoCache := os.Getenv("LLGOCACHE"); llgoCache != "" {
+		return llgoCache
+	}
+
+	// If not set, LLGOCACHE defaults to {UserCacheDir}/llgo/
+	userCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		panic("Can not get user cache dir: " + err.Error())
```
aofei:
```suggestion
		panic(fmt.Errorf("Can not get user cache dir: %w", err))
```

To preserve the original error in case an upper layer wants to recover from our panic, we use `%w` for wrapping instead of string concatenation. (See https://pkg.go.dev/fmt#Errorf)

compiler/internal/get/get.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func fetchModule(mod module.Version, flags []string) error {
+	args := []string{"get"}
+	args = append(args, flags...)
+	args = append(args, mod.String())
+
+	cmd := exec.Command("go", args...)
+	cmd.Stderr = os.Stderr
+	_, err := cmd.Output()
```
aofei: Is hiding stdout here intentional? If not, consider using `cmd.Run()` instead, and setting `cmd.Stdout = os.Stdout` above.

compiler/internal/mod/metadata.go
```diff
@@ -0,0 +1,104 @@
+package mod
+
+import (
+	"errors"
+	"path/filepath"
+)
+
+var (
+	remoteMetadataURL      = "https://llpkg.goplus.org/llpkgstore.json" // change only for testing
```
aofei: Replace this with `https://goplus.github.io/llpkg/llpkgstore.json`.

xushiwei:
Don't need a cache package. Just use a cache missing mechanism.

compiler/cmd/internal/get/get.go
```diff
@@ -32,5 +38,52 @@ func init() {
...
 
 func runCmd(cmd *base.Command, args []string) {
-	panic("todo")
+	if len(args) == 0 {
+		mockable.Exit(1)
+	}
+
+	flags := []string{}
+	flagEndIndex := -1
+	for idx, arg := range args {
```
xushiwei, line 47: use flag.Parse to parse command line arguments.

seri037: flag.Parse has been called in `main()`. we process the args it parsed here

seri037:
> Don't need a cache package. Just use a cache missing mechanism.

does it means that we should modify the function `metadata.update()`, and integrate all func. in `cache.go` to the `metadata.go`? Is it ok to mix file loading, file updating and file parsing functions all in one go file?

compiler/internal/get/get.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/3b07d0c87f01a73f19f9721dd000fc652c0d83b8

compiler/cmd/internal/get/get.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/326f1e89195c4b5c734c69ef2a81a3917bc0fb83

compiler/internal/env/env.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/801690bef6a9debb134300fdf9ba6453d774c971

compiler/internal/get/get.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/5449ef33364b0e1e9aa61bc3dbd974742cd2c624

compiler/internal/mod/metadata.go
seri037: done at https://github.com/goplus/llgo/pull/1038/commits/b5004f684f4004a3eccb5bebcae8cdb7638dc291

compiler/cmd/internal/get/get.go
aofei: I think the `args` passed in here are already all non-flag arguments, since they come from [`flag.Args()`](https://pkg.go.dev/flag#FlagSet.Args):

https://github.com/goplus/llgo/blob/3b9b71643cd360b4043606d0a3dc8207a51beeba/compiler/cmd/llgo/llgo.go#L59

PR #1041, opened 2025-03-28
feat: llgo list command with -m -versions -json flags

NEKO-CwC:
IMPORTANT: Because some of the llgo get code was manually merged locally, it is recommended to review after merging the pr [llgo get](https://github.com/goplus/llgo/pull/1038).

Moreover, the llgo list is currently in a semi-finished stage.

Temporarily, three flags have been implemented according to the following 思路.
0. llgo list supports two types of input: one is a string without a / and the other is a complete module path.
	1. When brief input is provided, such as cjson, the llpkgstore module's metadata section is used to obtain the latest cached content and structure of llpkgstore.json, to determine if it is a clib. If it is, the complete module path github.com/goplus/llpkg/{clib} is concatenated.
	2. If it is a complete module path, the logic below is directly entered.
1. The default behavior of llgo list is completely consistent with go list without any special flags.
2. When using llgo list -m
	1. The default behavior before is completely consistent with go list, including the conditions for use, and this verification can be simply implemented through code.
	2. If the verification is successful, you can enter the normal module list. At this point, you can concatenate the complete module path through environment variables, and if there is an llpkg.cfg in this path, identify it as LLPkg. When outputting, combine it with corresponding upstream information and the current version.
	3. If there is none, treat it as the default go module and fallback to go list for output.
3. When using llgo list -m -versions (there are two cases at this point, please write two different methods and put them in two files)
	1. Obtain the latest cached content and structure of llpkgstore.json through the metadata section of the llpkgstore module, and directly obtain the required version content by concatenating the methods here.
	2. For official llpkg, obtain the original file for upstream information acquisition by concatenating the address https://{module/path}/tree/{tag[vclib/version]}/{clib}/llpkg.cfg.
4. When using llgo list -json, simply inherit the flag to perform go list, then read the output into a structure, add the LLPkg field, and place the corresponding llpkg.cfg from the source code in it.


compiler/cmd/internal/list/list.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/goplus/llgo/compiler/cmd/internal/base"
+	"github.com/goplus/llgo/compiler/internal/list"
+	"github.com/goplus/llgo/compiler/internal/mockable"
+)
+
+// Cmd 定义list命令
+var Cmd = &base.Command{
+	UsageLine: "llgo list [-m] [-versions] [-json] [packages]",
+	Short:     "列出包、模块及其依赖的信息",
```
aofei: Only English is allowed in the code and comments.

NEKO-CwC: done in latest commit

PR #1103, opened 2025-05-09
add usage for llgo

tsingbx:


cmd/internal/clean/clean.go
```diff
@@ -35,6 +35,9 @@ func init() {
 }
 
 func runCmd(cmd *base.Command, args []string) {
+	defer func() {
```
xushiwei: Don't need this recover code. If we don't want to panic, change the original code.

PR #1104, opened 2025-05-09
xtool/cppkg:install in linux

luoliwoshang:
1. conan not in apt-get ' resource,so we could not use apt-get to install conan,and use pip install will got follow error,and the suggestion in the error message tips to use pipx.
```bash
> python3 -m pip install conan
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

--- 

user need install pipx,so when user not install pipx or brew notify error message with app manager.
```
root@259c7fd049f2:~/llgo/cmd/llgo# llgo cppkg install kspalaiologos/bzip3@1.5.1
From https://github.com/goplus/cppkg
 * branch              main       -> FETCH_HEAD
Already up to date.
panic: app managers not found: brew, pipx: executable file not found in $PATH
```

### DIFF
--- a/xtool/cppkg/command.go
+++ b/xtool/cppkg/command.go
@@ -17,6 +17,7 @@
 package cppkg
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"strings"
@@ -77,12 +78,14 @@ func (p *Tool) Get(quietInstall bool) (app string, err error) {
 }
 
 func (p *Tool) getAppManager() (amPath string, install []string, err error) {
+	var notFoundMgs []string
 	for _, install = range p.installs {
 		am := install[0]
 		if amPath, err = exec.LookPath(am); err == nil {
 			return
 		}
+		notFoundMgs = append(notFoundMgs, am)
 	}
-	err = ErrNotFound
+	err = fmt.Errorf("app managers not found: %s: %w", strings.Join(notFoundMgs, ", "), ErrNotFound)
 	return
 }
--- a/xtool/cppkg/conan.go
+++ b/xtool/cppkg/conan.go
@@ -37,7 +37,7 @@ import (
 
 var conanCmd = NewTool("conan", []string{
 	"brew install conan",
-	"apt-get install conan",
+	"pipx install conan",
 })
 
 type conandata struct {


xtool/cppkg/conan.go
```diff
@@ -255,6 +255,17 @@ func (p *Manager) conanfileDir(pkgPath, pkgFolder string) string {
 }
 
 func conanInstall(pkg, outDir, conanfileDir string, out io.Writer, options []string, flags int) (err error) {
+	quietInstall := flags&ToolQuietInstall != 0
+	app, err := conanCmd.Get(quietInstall)
+	if err != nil {
+		return
+	}
+
+	err = conanDetect(app, out)
```
xushiwei: should conan detect run every time?

xtool/cppkg/command.go
```diff
@@ -77,12 +78,14 @@ func (p *Tool) Get(quietInstall bool) (app string, err error) {
...
+	var notFoundMgs []string
 	for _, install = range p.installs {
 		am := install[0]
 		if amPath, err = exec.LookPath(am); err == nil {
 			return
 		}
+		notFoundMgs = append(notFoundMgs, am)
 	}
-	err = ErrNotFound
+	err = fmt.Errorf("app managers not found: %s: %w", strings.Join(notFoundMgs, ", "), ErrNotFound)
```
xushiwei, line 89: normally it's an unreachable path. so don't need have a kindly message.

luoliwoshang: On Linux, `pipx` might not be a very commonly used installation method, unlike brew or apt-get. So,I think a prompt like the one below could help reduce confusion during usage:
```
root@259c7fd049f2:~/llgo/cmd/llgo# llgo cppkg install kspalaiologos/bzip3@1.5.1  
From https://github.com/goplus/cppkg  
 * branch              main       -> FETCH_HEAD  
Already up to date.  
panic: app managers not found: brew, pipx: executable file not found in $PATH  
```

xtool/cppkg/conan.go
luoliwoshang: Since it's only necessary to detect the profile once after installing Conan, should we skip conan profile detect if a profile already exists (i.e., if `conan profile show` executes successfully)

```go
func conanDetect(app string, out io.Writer) (err error) {
	profileCmd := exec.Command(app, "profile", "show")
        // ....
	profileErr := profileCmd.Run()
	if profileErr == nil {
		return
	}
        // ...
```
or would you rather recommend letting the subsequent `conan install` step fail and prompt the user to run profile detect

Issue #1155, opened 2025-06-13
panic:Library not loaded: @rpath/libc++.1.dylib when LLGO_RPATH_CHANGE=on


luoliwoshang:

luoliwoshang:
`github.com/goplus/lib@v0.2.0/cpp/std/` parse LinkArgs to `-lc++`


before the exe replace these path to rpath,we can see the `/usr/lib/libc++.1.dylib` loaded

```bash
otool -L /var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/llgo-3090196538
/var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/llgo-3090196538:
        /opt/homebrew/opt/bdw-gc/lib/libgc.1.dylib (compatibility version 1.0.0, current version 1.5.4)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1900.180.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

but the  /usr/lib/libc++.1.dylib have not actual in this location

> In fact, these libraries don’t even exist on disk any more because they’re glommed into the dynamic linker shared cache

https://developer.apple.com/forums/thread/756370

so after replace to rpath of this exe,libc++ not loaded correct 
```bash 
llgo/internal/build on  main [$!?⇡] via 🐹 v1.24.1 
❯ otool -L /var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/llgo-3090196538
/var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/llgo-3090196538:
        @rpath/libgc.1.dylib (compatibility version 1.0.0, current version 1.5.4)
        @rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1900.180.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
(base) 
llgo/internal/build on  main [$!?⇡] via 🐹 v1.24.1 
❯  /var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/llgo-3090196538
dyld[99068]: Library not loaded: @rpath/libc++.1.dylib
[...]

PR #1156, opened 2025-06-13
fix(build):skip dylib path convert rpath which not found

luoliwoshang:
fix https://github.com/goplus/llgo/issues/1155

When the `LLGO_RPATH_CHANGE=true` flag is enabled on macOS, the compiler converts absolute dynamic library paths to be @rpath-relative. While this feature works correctly for third-party libraries, it causes a "library not found" error when the linked library is a system library (e.g., `libc++.1.dylib`).

 case at https://github.com/goplus/llgo/issues/1155,with a dyld error.
```bash
❯ llgo run .                 
ld: warning: ignoring duplicate libraries: '-lc++'
ld64.lld: warning: duplicate -rpath '/opt/homebrew/Cellar/bdw-gc/8.2.8/lib' ignored [--warn-duplicate-rpath]
ld64.lld: warning: duplicate -rpath '/opt/homebrew/Cellar/bdw-gc/8.2.8/lib' ignored [--warn-duplicate-rpath]
dyld[8499]: Library not loaded: @rpath/libc++.1.dylib
  Referenced from: <4C4C4440-5555-3144-A13C-496B4FB888AA> /Users/zhangzhiyang/go/bin/cpphello
  Reason: tried: '/opt/homebrew/Cellar/bdw-gc/8.2.8/lib/libc++.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/Cellar/bdw-gc/8.2.8/lib/libc++.1.dylib' (no such file), '/Users/zhangzhiyang/go/bin/libc++.1.dylib' (no such file), '/Users/zhangzhiyang/go/bin/../lib/libc++.1.dylib' (no such file), '/opt/homebrew/Cellar/bdw-gc/8.2.8/lib/libc++.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/Cellar/bdw-gc/8.2.8/lib/libc++.1.dylib' (no such file), '/Users/zhangzhiyang/go/bin/libc++.1.dylib' (no such file), '/Users/zhangzhiyang/go/bin/../lib/libc++.1.dylib' (no such file)
panic: signal: abort trap
```

The root cause is that system libraries on modern macOS do not exist as standalone files in their traditional paths (e.g.,  /usr/lib/libc++.1.dylib).

Although the initial executable correctly lists the system path for the dynamic library, as shown by otool -L:
```bash
otool -L /var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/llgo-3090196538
/var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/llgo-3090196538:
        /opt/homebrew/opt/bdw-gc/lib/libgc.1.dylib (compatibility version 1.0.0, current version 1.5.4)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1900.180.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```
In reality, the file is not physically present at that location. Instead, it resides in the `dyld` shared cache. As confirmed by Apple, these libraries are part of the dynamic linker shared cache:
> https://developer.apple.com/forums/thread/756370
> In fact, these libraries don’t even exist on disk any more because they’re glommed into the dynamic linker shared cache

converts /usr/lib/libc++.1.dylib to @rpath/libc++.1.dylib, the dynamic linker (dyld) attempts to find a physical file at the specified @rpath locations at runtime. This search fails, leading to the crash.

#### solution 

A possible method is to determine if a library path corresponds to a physical file on the filesystem. If a path does not resolve to an actual file (e.g., paths under `/usr/lib/libc++.dylib` on modern macOS), it should be treated as a system library located in the dyld shared cache. Consequently, its path should be left unmodified and not converted to be @rpath-relative.

### DIFF
--- a/cl/_testlibc/cppstr/in.go
+++ b/cl/_testlibc/cppstr/in.go
@@ -1,0 +1,12 @@
+package main
+
+import (
+	"github.com/goplus/lib/c"
+	"github.com/goplus/lib/cpp/std"
+)
+
+func main() {
+	s := std.Str("Hello world\n")
+	c.Printf(s.CStr())
+	print(s.Str(), s.Size(), "\n")
+}
--- a/cl/_testlibc/cppstr/out.ll
+++ b/cl/_testlibc/cppstr/out.ll
@@ -1,0 +1,48 @@
+; ModuleID = 'github.com/goplus/llgo/cl/_testlibc/cppstr'
+source_filename = "github.com/goplus/llgo/cl/_testlibc/cppstr"
+
+%"github.com/goplus/llgo/runtime/internal/runtime.String" = type { ptr, i64 }
+
+@"github.com/goplus/llgo/cl/_testlibc/cppstr.init$guard" = global i1 false, align 1
+@0 = private unnamed_addr constant [12 x i8] c"Hello world\0A", align 1
+@1 = private unnamed_addr constant [1 x i8] c"\0A", align 1
+
+define void @"github.com/goplus/llgo/cl/_testlibc/cppstr.init"() {
+_llgo_0:
+  %0 = load i1, ptr @"github.com/goplus/llgo/cl/_testlibc/cppstr.init$guard", align 1
+  br i1 %0, label %_llgo_2, label %_llgo_1
+
+_llgo_1:                                          ; preds = %_llgo_0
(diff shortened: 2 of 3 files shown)


xushiwei:
consider removing rpath support

PR #1160, opened 2025-06-18
internal/build:avoid dirty data read when clang link

luoliwoshang:


xushiwei:
1) pkg cache (.o + .a cache)
2) build hash (.o / tempdir + .a target name + dep modifications)
3) build proc conflict

PR #1161, opened 2025-06-19
build: -buildmode=c-archive

visualfc:
- support buildmode=c-archive
- ssa.BuildExpor for add export funcs first no phi instr call @pkg.init

```
// pkg/main.go
package main

//export add
func add(a,b int32) int32 {
    return a+b
}
func init() {
    println("load pkg init")
}
func main(){}
```
llgo build -buildmode=c-archive -o libmypkg.a .

```
// main.c
#include <stdio.h>
extern int add(int x, int y);
int main() {
    printf("hello world %d\n", add(100,200));
}
```
clang main.c -L./pkg -lmypkg -lgc -lffi -Wl,-dead_strip

### DIFF
--- a/cl/compile.go
+++ b/cl/compile.go
@@ -1048,6 +1048,7 @@ func NewPackageEx(prog llssa.Program, patches Patches, pkg *ssa.Package, files [
 		fn := ret.FuncOf(fnName)
 		if fn != nil {
 			fn.SetName(exportName)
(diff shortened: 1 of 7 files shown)


xushiwei:
1) .o + .a
2) init
3) buildmode
4) //export (allow in non-cgo)

PR #1187, opened 2025-07-14
[wip] esp32

luoliwoshang:
* need use the esp-clang to compile esp32

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -558,7 +558,6 @@ func linkMainPkg(ctx *context, pkg *packages.Package, pkgs []*aPackage, global l
 		}
 	})
 	entryLLFile, err := genMainModuleFile(ctx, llssa.PkgRuntime, pkg, needRuntime, needPyInit)
-	fmt.Println(entryLLFile)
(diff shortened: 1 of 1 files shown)


internal/build/build.go
```diff
@@ -1048,7 +1158,7 @@ func clFile(ctx *context, args []string, cFile, expFile string, procFile func(li
 		args = append(args, "-emit-llvm", "-S", "-o", llFile, "-c", cFile)
 	} else {
 		llFile += ".o"
-		args = append(args, "-o", llFile, "-c", cFile)
+		args = append(args, "-o", llFile, "-target", "xtensa-esp32-elf", "-c", cFile, "--target=xtensa-esp-unknown-elf", "-mcpu=esp32")
```
luoliwoshang, line 1147: NOTE: This may be specifying the target twice with potentially conflicting values:
  `-target xtensa-esp32-elf` and `--target=xtensa-esp-unknown-elf`. These might be
  duplicate flags that could cause unexpected behavior. We should consider using a single
  consistent target specification.

PR #1211, opened 2025-08-12
fix: set correct argc=0 in _start() function

luoliwoshang:
The original code used `icmp eq i32 0, 0` which always evaluates to true, resulting in argc=1:
```asm
 declare i32 @printf(i8*, ...)
  @fmt = private unnamed_addr constant [4 x i8] c"%d\0A\00", align 1

  define i32 @main() {
  entry:
    %argc_val = icmp eq i32 0, 0
    %argc = zext i1 %argc_val to i32
    %call = call i32 (i8*, ...) @printf(i8* getelementptr inbounds ([4 x i8], [4 x i8]* @fmt, i64 0, i64 0), i32 %argc)
    ret i32 0
  }
```
  Output: 1 (incorrect)

  After fix with add i32 0, 0, argc correctly equals 0.

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -719,8 +719,7 @@ call i32 @setvbuf(ptr %stderr_ptr, ptr null, i32 2, %size_t 0)
 	startDefine := `
 define weak void @_start() {
   ; argc = 0
-  %argc_val = icmp eq i32 0, 0
-  %argc = zext i1 %argc_val to i32
+  %argc = add i32 0, 0
   ; argv = null
   %argv = inttoptr i64 0 to i8**
   call i32 @main(i32 %argc, i8** %argv)


cpunion:
Fixing argv instead of argc. argc should be greater than or equal to 1, and argv[0] should be the executable path.

luoliwoshang:
> argc should be greater than or equal to 1, and argv[0] should be the executable path.

I found a few challenges:
  1. Executable path cannot be determined at compile time
  2. LLVM cannot dynamically retrieve this path - it maybe requires OS-specific syscalls
 
So i think the most reasonable solution might be setting argc = 1 and argv[0] to an empty string to at least comply with the C standard minimum requirements, rather than the current argc = 0, argv = null.

cpunion:
> I found a few challenges:
> 
> 1. Executable path cannot be determined at compile time
> 2. LLVM cannot dynamically retrieve this path - it maybe requires OS-specific syscalls
> 
> So i think the most reasonable solution might be setting argc = 1 and argv[0] to an empty string to at least comply with the C standard minimum requirements, rather than the current argc = 0, argv = null.

I agree with your solution.

PR #1214, opened 2025-08-12
internal/build:include target build-tags in Go package compilation

luoliwoshang:
The `hal` (Hardware Abstraction Layer) libraries depend on target-specific build-tags to compile the correct hardware drivers for each platform.


luoliwoshang:
@cpunion @xushiwei need review

cpunion (approved):

PR #1217, opened 2025-08-17
cl(feat):llgo.Asm

luoliwoshang:


_demo/asmcall/asmcall.go
```diff
@@ -0,0 +1,18 @@
...
+	"time"
+)
+
+//llgo:link asm llgo.asm
+func asm(instruction string) {}
+
+func main() {
+	start := time.Now()
+	for i := 0; i < 100000; i++ {
+		asm("nop")
```
luoliwoshang, line 14: NOTE:single `asm` call seems hardly to  make a directive function verification

cl/import.go
```diff
@@ -506,6 +506,9 @@ const (
 	llgoCgoCheckPointer = llgoCgoBase + 0x6
 	llgoCgoCgocall      = llgoCgoBase + 0x7
 
+	llgoAsmBase = llgoInstrBase + 0x40
```
xushiwei: asm don't have many instructions, so we don't need `llgoAsmBase`.

luoliwoshang: got it, we can determine whether the instruction corresponds to a complex or simple implementation based on the number of parameters.
remove `llgoAsmBase` at afee2beebe666f8783e60abe29650fb3da0d95d8

PR #1222, opened 2025-08-20
llgo flags -abi enable in build tags dev

visualfc:


internal/builddev/dev.go
```diff
@@ -0,0 +1,6 @@
+//go:build dev
+// +build dev
+
+package builddev
```
xushiwei: `buildenv` is better

Issue #1218, opened 2025-08-18
cl(feat): llgo.asm implement tinygo.AsmFull


luoliwoshang:
TinyGo provides two inline assembly interfaces:
- `device.Asm(asm string)` - Simple control instructions, no return value https://github.com/goplus/llgo/issues/1216
- `device.AsmFull(asm string, regs map[string]any) uintptr` - Advanced assembly, supports operands and return values

This document focuses on designing the implementation of the advanced assembly form `asmFull`:

#### Function Signature & Mapping

User declares the advanced assembly function and maps it to the `llgo.asm` compiler directive:

```go
// Advanced assembly function declaration (corresponding to TinyGo device.AsmFull)  
//go:linkname asmFull llgo.asm
func asmFull(instruction string, regs map[string]any) uintptr
```

**Compiler mapping**: Both `asm(string)` and `asmFull(string, map)` map to the same `llgo.asm` directive, distinguished by parameter count (1 vs 2 parameters).

#### Parameter Usage & Features

Design based on TinyGo, supporting:

**Parameters:**
- `instruction string` - Assembly instruction template with placeholder support
- `regs map[string]any` - Input operand mapping (key: placeholder name, value: actual value)
- `return uintptr` - Output register value when instruction contains output placeholder

**Supported features:**

1. **Placeholder system**
   - `{}` → `$0` (output operand, with return value)
   - `{name}` → `$1, $2...` (input operands, by order of appearance)

2. **Automatic constraint string generation**
   - `=&r` : Output to general register
   - `r` : Input from general register  

3. **Return value mechanism**
   - Returns `uintptr` when `{}` placeholder exists
   - Returns `void` when no output
   - **Design constraint**: Fixed single output return value (referencing TinyGo AsmFull's simplified design, balancing functionality and implementation complexity)

#### Usage

```go
func examples() {
    // Advanced assembly - with return value
    pc := asmFull("mov {}, sp", nil)                    // Read stack pointer
    
    // Advanced assembly - with input parameters
    asmFull("msr tpidr_el0, {value}", map[string]any{
        "value": 0x12345678,
    })
}
```

#### Expected LLVM IR Mapping Rules

Consistent interface and transformation approach with TinyGo, our expected transformation rules:

**Rule 1: Input parameters only** *(no return value, for control instructions)*
```go
asmFull("mov r0, {value}", map[string]any{
    "value": uint32(42),
})
```
Expected IR:
```llvm
[...]

PR #1224, opened 2025-08-20
cl(feat): llgo.asm implement tinygo.AsmFull

luoliwoshang:
resolved https://github.com/goplus/llgo/issues/1218

This PR fully adapts TinyGo's device.AsmFull implementation to provide LLGo with the same inline assembly capabilities, identical semantics, and constraint rules.
logic from [tinygo/compiler/inlineasm](https://github.com/tinygo-org/tinygo/blob/3869f76887feef6c444308e7e1531b7cac1bbd10/compiler/inlineasm.go#L46)

### usage

```go
//llgo:link asmFull llgo.asm
func asmFull(instruction string, regs map[string]any) uintptr { return 0 }
```
#### Parameters
- `instruction`: Assembly template with placeholders (`{}` for output, `{name}` for inputs)
- `regs`: Register mapping (`nil` for simple instructions, `map[string]any` for register values)

##### Example
* Control Instruction: CPU control instruction (no operation). Backward compatible with  simple `asm("nop")` from https://github.com/goplus/llgo/pull/1217.
```go
asmFull("nop", nil)
```
* Read Stack Pointer
```go
 pc := asmFull("mov {}, sp", nil)
```
* Memory Operation: Direct memory write
```go
addr := uintptr(unsafe.Pointer(&testVar))
  asmFull("str {value}, [{addr}]", map[string]any{
      "addr":  addr,
      "value": 43,
  })
```
also is aims to provide TinyGo memory operations for embedded systems. Enables precise hardware register access and memory control, similar to TinyGo's `device.AsmFull` used in GPIO and  peripheral programming.like follow operate 
```go
	riscv.AsmFull(
		"amoor.w {}, {mask}, ({reg})",
		map[string]interface{}{
			"mask": uint32(1 << pin),
			"reg":  uintptr(unsafe.Pointer(&kendryte.GPIOHS.FALL_IP.Reg)),
		})
```

#### tasks
- [x] return register
- [x] placeholder replace
- [x] llvm constraints
**test**
- [x] ir compare test
- [x] error case unit test
- [x] function test [_demo/asmfullcall/asmfullcall.go]
  - [x] macos
  - [x] linux
---

Maps used solely at compile time (like those for instruction signature compilation) currently still introduce runtime map dependencies, causing failures on embedded boards. we maybe can implement compile-time analysis to detect when programs only use compile-time function signatures, and exclude runtime capabilities in such cases.

### DIFF
--- a/cl/instr.go
+++ b/cl/instr.go
@@ -139,7 +139,8 @@ func (p *context) asm(b llssa.Builder, args []ssa.Value) (ret llssa.Expr) {
 				inputValues = append(inputValues, value)
 				constraints = append(constraints, "r")
 			} else {
-				// Pointer operands support was dropped, following TinyGo 0.23
+				// Pointer operands support was dropped, following TinyGo
+				// NOTE(tinygo): Memory references require a type starting with LLVM 14, probably as a preparation for opaque pointers.
 				panic(fmt.Sprintf("asm: unsupported type in inline assembly for operand: %s, only integer types are supported", name))
 			}
 		}


ssa/expr.go
```diff
@@ -286,6 +286,28 @@ func (b Builder) InlineAsm(instruction string) {
...
+
+	typs := make([]llvm.Type, len(exprs))
+	vals := make([]llvm.Value, len(exprs))
+	for i, expr := range exprs {
+		typs[i] = expr.Type.ll
+		vals[i] = expr.impl
+	}
+
+	typ := llvm.FunctionType(rtType, typs, false)
+	asm := llvm.InlineAsm(typ, instruction, constraints, true, false, llvm.InlineAsmDialectATT, false)
```
luoliwoshang: NOTE:`llvm.InlineAsm` return values are determined by fnType - can be void, single value, or multiple values via struct types.
This function may canbe better.

luoliwoshang: resolved at https://github.com/goplus/llgo/pull/1224/commits/8b9f2c63df3633dd44f1e18d1a76f6ebc608b70a

cl/instr.go
```diff
@@ -67,14 +70,89 @@ func cstr(b llssa.Builder, args []ssa.Value) (ret llssa.Expr) {
...
+	}
+
+	finalAsm = regexp.MustCompile(`\{[a-zA-Z]+\}`).ReplaceAllStringFunc(finalAsm, func(s string) string {
+		// TODO: skip strings like {r4} etc. that look like ARM push/pop
+		// instructions.
+		name := s[1 : len(s)-1]
+		value, ok := registers[name]
+		if !ok {
+			panic("asmFull: register not found: " + name)
+		}
```
luoliwoshang: need panic at `pointer type`
```go
	switch registers[name].Type().TypeKind() {
			case llvm.IntegerTypeKind:
				constraints = append(constraints, "r")
			case llvm.PointerTypeKind:
				// Memory references require a type starting with LLVM 14,
				// probably as a preparation for opaque pointers.
				err = b.makeError(instr.Pos(), "support for pointer operands was dropped in TinyGo 0.23")
     }
```

cl/instr.go
```diff
@@ -67,14 +70,89 @@ func cstr(b llssa.Builder, args []ssa.Value) (ret llssa.Expr) {
...
+	registerNumbers := map[string]int{}
+
+	if strings.Contains(finalAsm, "{}") {
+		finalAsm = strings.ReplaceAll(finalAsm, "{}", "$0")
+		constraints = append(constraints, "=&r")
+		registerNumbers[""] = 0
+		hasOutput = true
+	}
+
+	finalAsm = regexp.MustCompile(`\{[a-zA-Z]+\}`).ReplaceAllStringFunc(finalAsm, func(s string) string {
```
MeteorsLiu: NOTE(MeteorsLiu): `MustCompile` can be optimized by `once.Do()` to avoid compiling regexp repeatedly.

luoliwoshang: Done in [068253d](https://github.com/goplus/llgo/pull/1224/commits/068253dc415812cc07300fe6e947234107a29fed) but `once.Do()` maybe isn't necessary here since Go runtime already guarantees package-level variables are initialized only once during package initialization

cl/_testrt/asmfull/out.ll
```diff
@@ -0,0 +1,203 @@
...
+  ret void
+}
+
+define void @"github.com/goplus/llgo/cl/_testrt/asmfull.main"() {
+_llgo_0:
+  call void asm sideeffect "nop", ""()
+  %0 = load ptr, ptr @_llgo_string, align 8
+  %1 = load ptr, ptr @_llgo_any, align 8
+  %2 = load ptr, ptr @"map[_llgo_string]_llgo_any", align 8
+  %3 = call ptr @"github.com/goplus/llgo/runtime/internal/runtime.MakeMap"(ptr %2, i64 1)
```
cpunion, line 44: Remove runtime dependency in another PR

cl/instr.go
```diff
@@ -67,14 +72,93 @@ func cstr(b llssa.Builder, args []ssa.Value) (ret llssa.Expr) {
...
+
+	registers := make(map[string]llssa.Expr)
+	if registerMap, ok := args[1].(*ssa.MakeMap); ok {
+		referrers := registerMap.Referrers()
+		for _, r := range *referrers {
+			switch r := r.(type) {
+			case *ssa.DebugRef, *ssa.Call:
+				// ignore
+			case *ssa.MapUpdate:
+				if r.Block() != registerMap.Block() {
```
cpunion, line 98: Why?

_demo/asmfullcall/asmfullcall_linux.go
```diff
@@ -0,0 +1,30 @@
...
+package main
+
+import "unsafe"
+
+func verify() {
+	// 0 output & 0 input
+	asmFull("nop", nil)
+
+	// 0 output & 1 input with memory address
+	addr := uintptr(unsafe.Pointer(&testVar))
```
cpunion, line 12: Why pointer type is not supported?

cl/instr.go
```diff
@@ -67,14 +72,93 @@ func cstr(b llssa.Builder, args []ssa.Value) (ret llssa.Expr) {
...
+		log.Printf("asm: %q -> %q, constraints: %q", asmString, finalAsm, constraintStr)
+	}
+
+	var retType llssa.Type
+	if hasOutput {
+		retType = b.Prog.Uintptr()
+	} else {
+		retType = b.Prog.Void()
+	}
+
```
cpunion, line 160: Is retType matched with the func return type?

cl/instr.go
luoliwoshang: this panic logic is fully consistent with current TinyGo behavior, and tracing back to the earliest commit of this code (https://github.com/tinygo-org/tinygo/commit/392bba8394ece9e8b90ee2d6e82590b1492bc62d), no reasoning was provided. 
However, I believe the fundamental technical reason is that **LLVM inline assembly requires all call information to be determined at compile time** - parameter types, constraint strings, and parameter counts cannot change dynamically at runtime.

```llvm
call i64 asm sideeffect "mov $0, ${1}", "=&r,r"(i64 42)
```

This is essentially the same requirement as any function call needing determined function signatures and parameter types at compile time.

If cross-basic-block map construction were allowed, the compiler couldn't determine what parameters are needed for the call:

```go
regs := make(map[string]interface{})
  if condition {
      regs["value"] = 42
  }
res1 := asmFull("mov {}, {value}", regs)  // Compiler can't determine if value exists 
```

luoliwoshang: at this `case *ssa.MapUpdate:` we actually can't know if this value needs collecting at compile time.

cl/instr.go
luoliwoshang: fixed at [49b9b92](https://github.com/goplus/llgo/pull/1224/commits/49b9b92790ba6541776720a70a3f6bf2a67955b5), Now when hasOutput is false, we return a valid uintptr(0) value that matches the Go function signature, rather than returning a void-typed inline assembly call.

_demo/asmfullcall/asmfullcall_linux.go
luoliwoshang: Current logic is fully corespoding with `Tinygo`,so panic with same case.
And the reason TinyGo currently doesn't support pointer operands in inline assembly is due to LLVM's transition from typed pointers to opaque pointers starting with LLVM 14 (https://llvm.org/docs/OpaquePointers.html). This change requires inline assembly to explicitly declare pointer element types using the `elementtype` attribute (https://reviews.llvm.org/D116531). Rather than  implementing the additional complexity, TinyGo chose to remove pointer support entirely (https://github.com/tinygo-org/tinygo/commit/cad6a57077c7887025ba532a03f41d6ad78daa72).

However, I believe it's possible to add back this pointer type support. As TinyGo's comment indicates, it would require modifying the go-llvm API. I think we should make this functionality available first, and then support this feature in a separate PR (after all, going from `uintptr(unsafe.Pointer(&val))` to `&val` wouldn't introduce breaking changes).

cpunion (approved):

luoliwoshang:
@xushiwei need review

PR #1163, opened 2025-06-23
internal/build:support relocatable lib

luoliwoshang:
fixed https://github.com/goplus/llgo/issues/1135

- [x] LLGO_FULL_RPATH to controll every comile process's lib path set to RPATH
- [x] test at https://github.com/Homebrew/homebrew-core/pull/227010

### reason

library with conan with generate install_name directly with `@rpath` and the go is not add rpath for these dylib.
```bash
llgo/d on  internal/build/rpath [$!?] via C v17.0.0-clang 
❯ otool -D /opt/homebrew/Cellar/cjson/1.7.18/lib/libcjson_utils.dylib
/opt/homebrew/Cellar/cjson/1.7.18/lib/libcjson_utils.dylib:
/opt/homebrew/opt/cjson/lib/libcjson_utils.1.dylib
(base) 
llgo/d on  internal/build/rpath [$!?] via C v17.0.0-clang 
❯ otool -D /Users/zhangzhiyang/.conan2/p/b/cjsond2bd82ff69524/p/lib/libcjson_utils.dylib 
/Users/zhangzhiyang/.conan2/p/b/cjsond2bd82ff69524/p/lib/libcjson_utils.dylib:
@rpath/libcjson_utils.1.dylib
```
### resolution

Treat every link-time library search path, specified by the -L parameter, as a runtime search path as well.This is to ensure the final executable can locate libraries with a relocatable install_name (e.g., "@rpath/libfoo.dylib") at runtime.

current run normally

PR #1237, opened 2025-08-25
[WIP]feat:support isolated python runtime for building

1351914167:
- New module `internal/pyenv`
- Purpose: during `llgo build`, automatically download a Python environment into `LLGoCache`, pip corresponding site-packages, and set up the runtime.

Current issues:
1. Only supported under `llgo run` and `llgo build`. When running the binary, the program falls back to the system’s default Python.
2. Only macOS Python builds are supported. On Ubuntu, Go–Python programs will panic with “todo” instead of using the system’s default Python. Therefore, all Ubuntu jobs in CI failing is expected.


internal/pyenv/pybuild.go
```diff
@@ -0,0 +1,320 @@
...
+		_ = os.Setenv("PKG_CONFIG_PATH", pkgcfg)
+	} else {
+		parts := strings.Split(pcp, string(os.PathListSeparator))
+		found := false
+		for _, p := range parts {
+			if p == pkgcfg {
+				found = true
+				break
+			}
+		}
```
MeteorsLiu: It's better to replace with `slices.Contains`

internal/pyenv/pybuild.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func EnsurePcRpath(pyHome string) error {
+	pc := filepath.Join(pyHome, "lib", "pkgconfig", "python3-embed.pc")
+	b, err := os.ReadFile(pc)
+	if err != nil {
+		return err
+	}
+	lines := strings.Split(string(b), "\n")
+	rpath := "-Wl,-rpath," + filepath.Join(pyHome, "lib")
```
MeteorsLiu, line 186: It seems  that's unnecessary,  #1163 will set up `-rpath` automatically.

1351914167: Done in [5643951](https://github.com/goplus/llgo/pull/1237/commits/564395164eca37d67f8b0d44afee7f864c014c4d), I keep FindPythonRpath to dynamically resolve the local Python library path.

internal/pyenv/pybuild.go
1351914167: Done in [3f9c97b](https://github.com/goplus/llgo/pull/1237/commits/3f9c97b676156aa1e0cc8f48b64d8896a1e3d85f).

PR #1242, opened 2025-08-27
[WIP] internal/cabi: call optimization

visualfc:
call optimization:  check IR `alloca/load/store` =>  `direct ptr` / `bitcast` / `memcpy` 

transform func body

func params
- [x] AttrPointer (byval) : check alloca ptr
- [x] AttrWidthType :  bitcast,  check alloca ptr
- [x] AttrWidthType2:  bitcast, check alloca ptr
- [x] AttrExtract: none

func results
- [x] AttrPointer sret : check memcpy
- [x] AttrWidthType : check alloca ptr, bitcast
- [x] AttrWidthType2: check alloca ptr, bitcast
- [x] AttrExtract: none

transform call instr
- [x] AttrPointer sret : check direct ptr
- [x] AttrPointer (byval) : check direct ptr or memcpy( unsupport byval, arm/arm64 )
- [ ] AttrWidthType
- [ ] AttrWidthType2
- [ ] AttrExtract

### DIFF
--- a/internal/cabi/cabi.go
+++ b/internal/cabi/cabi.go
@@ -113,7 +113,7 @@ func (p *Transformer) TransformModule(path string, m llvm.Module) {
 		}
 	}
 	for _, call := range callInstrs {
-		p.transformCallInstr(ctx, call.call, call.fn)
+		p.transformCallInstr(m, ctx, call.call, call.fn)
 	}
 	for _, fn := range fns {
 		p.transformFunc(m, fn)
@@ -368,9 +368,8 @@ func (p *Transformer) transformFuncBody(m llvm.Module, ctx llvm.Context, info *F
 			// call void @llvm.memset(ptr %2, i8 0, i64 36, i1 false)
 			// store %typ %1, ptr %2, align 4
 			nv = b.CreateLoad(ti.Type, params[index], "")
-			if p.sys.SupportByVal() {
-				replaceParamAlloc(fn.Param(i), params[index])
-			}
+			// replace %0 to %2
+			replaceParamAlloc(fn.Param(i), params[index])
 		case AttrWidthType:
 			iptr := llvm.CreateAlloca(b, ti.Type1)
 			b.CreateStore(params[index], iptr)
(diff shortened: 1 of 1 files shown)


internal/cabi/cabi.go
```diff
@@ -613,3 +629,50 @@ func (p *Transformer) transformCallbackFunc(m llvm.Module, fn llvm.Value) (wrap
...
+		sz = llvm.ConstInt(ctx.Int64Type(), uint64(size), false)
+	}
+	return b.CreateCall(memcpy.GlobalValueType(), memcpy, []llvm.Value{
+		dst, src, sz, llvm.ConstInt(ctx.Int1Type(), 0, false),
+	}, "")
+}
+
+const (
+	memcpy64 = "llvm.memcpy.p0i8.p0i8.i64"
+	memcpy32 = "llvm.memcpy.p0i8.p0i8.i32"
```
cpunion: Reference from https://llvm.org/docs/LangRef.html#llvm-memcpy-intrinsic :
```
This is an overloaded intrinsic. You can use llvm.memcpy on any integer bit width and for different address spaces. Not all targets support all bit widths however.
```

See simular usage:
https://github.com/goplus/llgo/blob/cf2d1ef9ca2f56bc6fda5d083d90507bfbd0863a/ssa/memory.go#L260-L264

Just use `llvm.memcpy` as func name (correct function type is needed).

Maybe `cFunc` is not a good usage, need a better intrinsic wrapper.

internal/cabi/cabi.go
```diff
@@ -31,23 +31,27 @@ func NewTransformer(prog ssa.Program, mode Mode) *Transformer {
...
-	GOARCH string
-	sys    TypeInfoSys
-	mode   Mode
+	prog    ssa.Program
+	td      llvm.TargetData
+	GOOS    string
+	GOARCH  string
+	sys     TypeInfoSys
+	mode    Mode
+	is32Bit bool
```
cpunion: Maybe add `func (p *Program) SizeT() Type`, and then don't need `is32Bit` field?

internal/cabi/cabi.go
visualfc: fixed

internal/cabi/cabi.go
visualfc: use ctx.IntType(p.prog.PointerSize() * 8)

PR #1245, opened 2025-08-28
internal/target:read targets consist use LLGoROOT

luoliwoshang:
Replace hardcoded runtime.Caller() path calculation with env.LLGoROOT() in targets resolver for consistency with runtime directory resolution.
This aligns targets directory resolution with the existing runtime directory resolution strategy.

### DIFF
--- a/internal/targets/resolver.go
+++ b/internal/targets/resolver.go
@@ -3,7 +3,8 @@ package targets
 import (
 	"fmt"
 	"path/filepath"
-	"runtime"
+
+	"github.com/goplus/llgo/internal/env"
 )
 
 // Resolver provides high-level interface for target configuration resolution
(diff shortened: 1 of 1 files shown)


internal/targets/resolver.go
```diff
@@ -20,9 +21,7 @@ func NewResolver(targetsDir string) *Resolver {
 
 // NewDefaultResolver creates a resolver with default targets directory
 func NewDefaultResolver() *Resolver {
-	// Assume targets directory is relative to this package
-	_, filename, _, _ := runtime.Caller(0)
-	projectRoot := filepath.Dir(filepath.Dir(filepath.Dir(filename)))
+	projectRoot := env.LLGoROOT()
```
cpunion: Rename to llgoRoot

cpunion (changes requested):

internal/targets/resolver.go
luoliwoshang: Done

PR #1268, opened 2025-09-05
fix(runtime): go runtime when building in baremetal environment

MeteorsLiu:
TODO: sigsetjmp(defer)


cpunion (approved):

runtime/internal/clite/c.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	LLGoPackage = "decl"
+	LLGoPackage = "link"
```
luoliwoshang: Is this update to avoid some runtime init?

runtime/internal/clite/stdio_baremetal.go
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package c
+
+import _ "unsafe"
+
+// initialized by z_print because LLGoPackage skips calling init()
```
luoliwoshang: maybe is the `LLGoPackage = "link"` skips calling init()?
```suggestion
// initialized by z_print because LLGoPackage = "link" skips calling init()
```

runtime/internal/runtime/z_print.go
```diff
@@ -22,6 +22,18 @@ import (
...
+	if c.Stdin == nil {
+		c.Stdin = c.Fopen(c.Str("/dev/stdin"), c.Str("r"))
+	}
+	if c.Stdout == nil {
+		c.Stdout = c.Fopen(c.Str("/dev/stdout"), c.Str("w"))
+	}
+	if c.Stderr == nil {
+		c.Stderr = c.Stdout
+	}
+}
```
luoliwoshang: The init() function in `z_print.go` appears to be specifically designed to initialize the stdio variables from `stdio_baremetal.go`. Consider creating a dedicated `z_print_baremetal.go` file with `//go:build baremetal` to make this relationship clearer, and add `//go:build !baremetal` to the existing file.

runtime/internal/runtime/z_baremetal.go
```diff
@@ -0,0 +1,19 @@
...
+
+package runtime
+
+import c "github.com/goplus/llgo/runtime/internal/clite"
+
+var (
+	printFormatPrefixInt  = c.Str("%ld")
+	printFormatPrefixUInt = c.Str("%lu")
+	printFormatPrefixHex  = c.Str("%lx")
+)
```
MeteorsLiu, line 11: NOTE: some libc may not support `%lld`

runtime/internal/runtime/z_print.go
MeteorsLiu: It's not only for `z_print_baremetal`. `Stdout` / `Stderr` are exported symbols which means it should may be used by another, this codes ensure `Stdout` / `Stderr` not be `nil`.

runtime/internal/clite/stdio_baremetal.go
MeteorsLiu: No. When `LLGoPackage = "decl"`,  skips executing `init()` as well, the commets are right.

runtime/internal/clite/c.go
MeteorsLiu: When `LLGoPackage = "decl"`, `Stdout` / `Stderr` should not be `nil` which will cause symbols not found. However, in `newlib-esp32`, we cannot link `stdout` / `stderr`, because they are macros. So we have to make `LLGoPackage = "link"` insteand of `LLGoPackage = "decl"`.

MeteorsLiu: It seems `LLGoPackage = true` will be better.
When `LLGoPackage = true`, `init()` can be executed, while we can use `llgo:link` as well.

runtime/internal/runtime/z_print.go
MeteorsLiu: Remove theses codes in https://github.com/goplus/llgo/pull/1268/commits/0bbdb6c2046a7da49adb28e965f948f9bd4e9c01

runtime/internal/clite/c.go
luoliwoshang: LGTM

luoliwoshang (approved):

PR #1301, opened 2025-09-15
llgo/embed:compile with `Oz`  & link with `-s` for embed target to reduce size

luoliwoshang:
fixed https://github.com/goplus/llgo/issues/1312

before
```bash
2025-09-16T08:14:33.7708327Z ld.lld: error: section 'text' will not fit in region 'iram_seg': overflowed by 13157 bytes
```
with current pr 190KB -> 86KB
```bash
lib/_demo/emptycheck on  task3/machine [?] via 🐹 v1.24.6 took 7s 
❯ ls -lh ./oz-s.elf                                             
-rwxr-xr-x@ 1 zhangzhiyang  staff    86K Sep 18 15:23 ./oz-s.elf
```

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -834,7 +834,6 @@ func linkObjFiles(ctx *context, app string, objFiles, linkArgs []string, verbose
 	}
 
 	buildArgs := []string{"-o", app}
-	buildArgs = append(buildArgs, "-s")
 	buildArgs = append(buildArgs, linkArgs...)
 
 	// Add build mode specific linker arguments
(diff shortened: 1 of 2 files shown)


targets/xtensa.json
```diff
@@ -18,7 +18,8 @@
 		"-fno-unwind-tables",
 		"-fno-asynchronous-unwind-tables",
 		"-ffunction-sections",
-		"-fdata-sections"
+		"-fdata-sections",
+		"-Os"
```
MeteorsLiu: In summary, this is required not only for `Xtensa`, but for all bare-metal targets. Consider to add the `-Os` option for them?

luoliwoshang: Done,current is for every embed target

MeteorsLiu (approved):

PR #1310, opened 2025-09-16
fix: disable buffers

MeteorsLiu:


runtime/internal/clite/stdio_baremetal.go
```diff
@@ -25,3 +25,16 @@ import (
...
+
+const (
+	_IONBF = 2 // No buffering - immediate output
+)
+
+func init() {
+	// Disable buffering for baremetal targets to ensure immediate output
+	setvbuf(Stdout, nil, _IONBF, 0)
+	setvbuf(Stdin, nil, _IONBF, 0)
+}
```
cpunion, line 40: This is just a workaround, not a solution. Why not expose to `github.com/goplus/lib/c.Setvbuf`?

MeteorsLiu:
Closed because not a good solution

PR #1323, opened 2025-09-24
runtime: math/big use math_big_pure_go

visualfc:
use math_big_pure_go for full support math/big


internal/build/build.go
```diff
@@ -216,7 +216,7 @@ func Do(args []string, conf *Config) ([]Package, error) {
 
 	verbose := conf.Verbose
 	patterns := args
-	tags := "llgo"
+	tags := "llgo,math_big_pure_go"
```
xushiwei, line 219: don't need to define `math_big_pure_go` tag

PR #1350, opened 2025-10-16
fix: override go/build.defaultContext() to use gc compiler

luoliwoshang:
Based on https://github.com/goplus/llgo/pull/1349, the conversation became too long for the agent to reply, so I opened this PR to continue the work.

### DIFF
--- a/_demo/go/gobuild/demo.go
+++ b/_demo/go/gobuild/demo.go
@@ -4,31 +4,102 @@ import (
 	"fmt"
 	"go/build"
 	"runtime"
+	"strings"
 )
 
 func main() {
-	fmt.Printf("runtime.Compiler = %q\n", runtime.Compiler)
+	fmt.Println("=== Testing go/build package with llgo ===")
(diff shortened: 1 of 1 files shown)


runtime/internal/lib/runtime/symtab.go
```diff
@@ -140,6 +140,10 @@ type Func struct {
 	opaque struct{} // unexported field to disallow conversions
 }
 
+func (f *Func) Name() string {
```
luoliwoshang, line 143: NOTE:keep consist with `runtime/_overlay/runtime/runtime.go` avoid follow
```bash
ld64.lld: error: undefined symbol: runtime.(*Func).Name
>>> referenced by /Users/runner/Library/Caches/go-build/7b/7b7511c7aa2fcfba0c66cb5102aff968d6dab03c9b7a877ebe93895d51fc94aa-d.o:(symbol internal/bisect.printStack+0x228)
```

Issue #1357, opened 2025-10-17
`go/types` scope.Insert got `segmentation fault`


luoliwoshang:
```go
package main

import (
	"fmt"
	"go/types"
)

func main() {
	fmt.Println("=== Scope Insert Test ===\n")

	// Create a scope
	scope := types.NewScope(nil, 0, 0, "test")

	// Test 1: Successful insert
	obj1 := types.NewVar(0, nil, "x", types.Typ[types.Int])
	if alt := scope.Insert(obj1); alt != nil {
		panic(fmt.Sprintf("unexpected: insert failed with conflict: %v", alt))
	}
	fmt.Printf("✓ Insert successful: %s\n", obj1.Name())

	// Test 2: Duplicate insert (will fail)
	obj2 := types.NewVar(0, nil, "x", types.Typ[types.String])
	if alt := scope.Insert(obj2); alt != nil {
		fmt.Printf("✓ Insert failed as expected: %s already exists\n", alt.Name())
	} else {
		panic("unexpected: duplicate insert should have failed")
	}

	// Test 3: Insert different name
	obj3 := types.NewVar(0, nil, "y", types.Typ[types.String])
	if alt := scope.Insert(obj3); alt != nil {
		panic(fmt.Sprintf("unexpected: insert failed with conflict: %v", alt))
	}
	fmt.Printf("✓ Insert successful: %s\n", obj3.Name())

	// View results
	fmt.Printf("\nScope contains %d objects\n", scope.Len())

	// Verify expected count
	expectedCount := 2 // x and y
	if scope.Len() != expectedCount {
		panic(fmt.Sprintf("unexpected: scope should contain %d objects, but has %d", expectedCount, scope.Len()))
	}
	fmt.Printf("✓ Scope length check passed\n")

	fmt.Println("\nAll objects:")
	for _, name := range scope.Names() {
		obj := scope.Lookup(name)
		fmt.Printf("  - %s: %v\n", name, obj.Type())
	}
}
```

expect got
```bash
=== Scope Insert Test ===

✓ Insert successful: x
✓ Insert failed as expected: x already exists
✓ Insert successful: y

Scope contains 2 objects
✓ Scope length check passed

All objects:
  - x: int
  - y: string
```
but got follow with `llgo run`
```bash
=== Scope Insert Test ===

signal: segmentation fault
```

luoliwoshang:
@xgopilot try to reproduce this problem! if you can reproduce,invest it!

[...]

PR #1387, opened 2025-11-06
fix(runtime): segmentation fault when calling interface private methods cross-package

### LINKED ISSUE #1370: Segmentation Fault When Calling Interface Private Methods Cross-Package
## Problem Description

In llgo, when a concrete type pointer is converted to an interface with private methods in a package different from the interface definition package, the interface metadata's `PkgPath` is incorrectly set to the current compilation package instead of the interface definition package. This causes the runtime to only fill exported methods in the itab, leaving private method slots as NULL (0x0), which leads to segmentation faults when calling these private methods.

**Impact Scope**:
- ✅ Go standard library interfaces:
  - `go/types.Object` (private method `setParent()`)
  - `go/ast.Expr` (private method `exprNode()`)
- ✅ User-defined packages (cross-package interface conversion)

## Trigger Conditions

All of the following conditions must be met:

1. ✅ Interface contains **private methods** (unexported)
2. ✅ Function returns **concrete type pointer** (e.g., `*Rectangle`, `*types.Var`)
3. ✅ Interface conversion occurs in **caller package** (not the interface definition package)

### Case 1: Standard Library - go/types.Object

```go
package main

import (
	"go/types"
)

func main() {
	scope := types.NewScope(nil, 0, 0, "test")
	obj := types.NewVar(0, nil, "x", types.Typ[types.Int])

	scope.Insert(obj)  // ← Segmentation fault
}
```

**Error Output**:
```
=== Scope Insert Test ===

Inserting object into scope...
signal: segmentation fault
```

**Crash Location**: `go/types/scope.go:105` - `obj.setParent(s)`

### Case 2: Standard Library - go/ast.Expr

```go
package main

import (
	"fmt"
	"go/ast"
	"go/token"
)

func main() {
	ident := &ast.Ident{
		NamePos: token.Pos(1),
		Name:    "foo",
	}

	var expr ast.Expr = ident  // ← Interface conversion in main package
	fmt.Printf("Position: %d\n", expr.Pos())
}
```

The `go/ast.Expr` interface has a private method `exprNode()`, causing the same issue.

### Case 3: Custom Package - geometry.Shape

**Definition Package** (`geometry/geometry.go`):
```go
package geometry

type Shape interface {
	Area() float64
	validate() bool  // Private method
	setID(int)       // Private method
}

type Rectangle struct {
	Width, Height float64
	id int
}

func (r *Rectangle) Area() float64      { return r.Width * r.Height }
func (r *Rectangle) validate() bool     { return r.Width > 0 && r.Height > 0 }
func (r *Rectangle) setID(id int)       { r.id = id }

// Returns concrete type pointer
func NewRectangle(width, height float64) *Rectangle {
	return &Rectangle{Width: width, Height: height}
}

func RegisterShape(s Shape, id int) error {
	s.setID(id)  // ← Triggers segfault
	return nil
}
```

**Caller** (`main.go`):
```go
package main

import "geometry"

func main() {
	rect := geometry.NewRectangle(5.0, 3.0)
	geometry.RegisterShape(rect, 42)  // ← Segmentation fault
}
```

### Why Doesn't Returning Interface Type Cause Issues?

| Scenario | Return Type | Interface Conversion Location | Triggers Bug |
|----------|------------|-------------------------------|--------------|
| **Case A** | `func New() Shape` | Inside definition package | ❌ No |
| **Case B** | `func New() *Rectangle` | In caller package | ✅ Yes |

**Case A** (Safe): Interface conversion happens in the definition package, generating correct `@"geometry.iface$..."` with `PkgPath = "geometry"`.

**Case B** (Bug): Concrete type pointer is returned, conversion happens in caller package, generating `@"command-line-arguments.iface$..."` with wrong `PkgPath = "command-line-arguments"`.

### Step 1: lldb Debugging

Using lldb to debug the `go/types.Object` case:

```
* thread #1, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x1002cd228 test_insert`go/types.(*Scope).Insert + 236

register read:
    pc = 0x0000000000000000  ← Jump to address 0
    lr = 0x1002cd228         ← Called from Insert +236
```

**Assembly Analysis** (using `otool -tV`):

```assembly
# s.insert(name, obj)
1002cd1e8  bl   "_go/types.(*Scope).insert"

# if obj.Parent() == nil
1002cd1f4  bl   "_...IfacePtrData"        # Extract interface.data
1002cd1f8  ldr  x8, [sp, #0x10]           # Read interface (itab)
1002cd1fc  ldr  x8, [x8, #0x30]           # Read Parent() pointer (offset 0x30)
1002cd200  blr  x8                        # Call Parent() ✅ Success
1002cd204  cbnz x0, 0x1002cd22c           # Check return value

# obj.setParent(s)
1002cd214  bl   "_...IfacePtrData"        # Extract interface.data
1002cd218  ldr  x8, [sp, #0x10]           # Read interface (itab)
1002cd220  ldr  x8, [x8, #0x88]           # Read setParent() pointer (offset 0x88)
                                          # ← Reads 0x0 here!
1002cd224  blr  x8                        # Call setParent() ❌ Crash!
```

**Key Finding**:

| Method | itab Offset | Value Read | Result |
|--------|------------|-----------|--------|
| `Parent()` | 0x30 (48) | Valid function address | ✅ Success |
| `setParent()` | 0x88 (136) | **0x0 (NULL)** | ❌ Crash |

The itab has a valid function pointer for `Parent()` but NULL for `setParent()`.

### Step 2: Adding Debug Output to Runtime

To understand why `setParent()` is not filled in itab, we added debug output to `runtime/internal/runtime/z_face.go`:

```go
func NewItab(inter *InterfaceType, typ *Type) *Itab {
    c.Printf(c.Str(" inter.PkgPath_: %s\n"), c.AllocCStr(inter.PkgPath_))
    // ...
}

func methods(u *abi.UncommonType, from string) []abi.Method {
    c.Printf(c.Str("  methods, from=%s, u.PkgPath_=%s\n"),
             c.AllocCStr(from), c.AllocCStr(u.PkgPath_))
    if u.PkgPath_ == from {
        return u.Methods()           // Return all methods (including private)
    }
    return u.ExportedMethods()       // Return exported methods only
}
```

**Re-run the go/types test**:

```
=== Scope Insert Test ===

Inserting object into scope...
 inter.PkgPath_: command-line-arguments  ← 🔴 Wrong! Should be "go/types"
  methods, from=command-line-arguments, u.PkgPath_=go/types
Method Itab Process end
signal: segmentation fault
```

**Critical Discovery**: The interface metadata's `PkgPath` is incorrectly set to `"command-line-arguments"` instead of `"go/types"`!

### Why This Causes the Crash

**Runtime itab construction fails** (`runtime/internal/runtime/z_face.go`):

```go
func NewItab(inter *InterfaceType, typ *Type) *Itab {
    // ...
    mthds := methods(u, inter.PkgPath_)  // inter.PkgPath_ = "command-line-arguments"
    // ...
}

func methods(u *abi.UncommonType, from string) []abi.Method {
    if u.PkgPath_ == from {              // "go/types" != "command-line-arguments"
        return u.Methods()               // ← Doesn't execute (would return all methods)
    }
    return u.ExportedMethods()           // ← Executes! Returns only exported methods
}
```

**Result**:
- `methods()` returns only exported methods (`Exported`, `Id`, `Name`, `Parent`, etc.)
- Private method `setParent` is filtered out
- `findMethod()` can't find `setParent` → returns `nil`
- itab slot for `setParent` remains `0x0`
- Compiler generates code to call `setParent` at offset 0x88 in itab
- Runtime reads from offset 0x88 → gets `0x0`
- Execute `blr x8` (x8 = 0x0) → jump to address 0 → **crash**

### Compiler vs Runtime Inconsistency

**At compile-time**, the compiler calculates the itab offset for `setParent` based on the interface method order (`ssa/interface.go:56-82`):

```go
func iMethodOf(rawIntf *types.Interface, name string) int {
    n := rawIntf.NumMethods()
    for i := 0; i < n; i++ {
        m := rawIntf.Method(i)
        if m.Name() == name {
            return i  // setParent → index 14
        }
    }
    return -1
}

func (b Builder) Imethod(intf Expr, method *types.Func) Expr {
    prog := b.Prog
    rawIntf := intf.raw.Type.Underlying().(*types.Interface)
    i := iMethodOf(rawIntf, method.Name())  // i = 14 for setParent
    // ...
    pfn := b.Advance(itab, prog.IntVal(uint64(i+3), prog.Int()))
    // Offset = (14+3) × 8 = 136 = 0x88
    fn := b.Load(pfn)  // Generates: ldr x8, [itab, #0x88]
    // ...
}
```

**Compiler assumes**: All 17 methods of `go/types.Object` will be in itab, so `setParent` (method #14) is at offset 0x88.

**At runtime**, due to wrong PkgPath, only 9 exported methods are filled, leaving `setParent`'s slot empty.

**Conflict**:
- Compiler generates: `ldr x8, [itab, #0x88]` (expects setParent at index 14)
- Runtime fills: Only indices 0-8 with exported methods
- Result: Compiler reads index 14 → gets 0x0 → crash

### Execution Flow Summary

```
1. main package: obj := types.NewVar(...)
   Returns *types.Var (concrete type pointer)
   ↓
2. main package: scope.Insert(obj)
   Needs to convert *types.Var → types.Object interface
   ↓
3. Compiler generates interface metadata (BUG HERE!)
   PkgPath set to "command-line-arguments" instead of "go/types"
   ↓
4. Runtime calls NewItab(Object interface, *Var type)
   inter.PkgPath_ = "command-line-arguments" ❌
   u.PkgPath_ = "go/types" ✅
   ↓
5. methods() detects package mismatch
   Returns only ExportedMethods()
   ↓
6. findMethod() can't find setParent
   itab slot for setParent = 0x0
   ↓
7. obj.setParent(s) reads NULL from itab
   blr x8 (x8=0) → crash
```

**Why This is a Visibility Issue**:

Go's package visibility mechanism in `methods()`:
- **Same package** (`u.PkgPath_ == from`): Returns all methods including private ones
- **Cross package** (`u.PkgPath_ != from`): Returns only exported methods

When `inter.PkgPath_` is wrong, the runtime thinks it's a cross-package call and refuses to fill private methods in itab, even though the concrete type is from the same package as the interface.

## Related Files

- **Bug location**: `ssa/abi/abi.go:279` - `InterfaceName()` function
- **Runtime logic**: `runtime/internal/runtime/z_face.go` - `NewItab()` and `methods()` functions
- **Detailed investigation**: `.issue/scope-insert-segfault/docs/INVESTIGATION.md`

luoliwoshang:
Fixes #1370 #1357

  ## Problem

  LLGo previously used an optimization to pre-filter methods before itab matching:

  ```go
  mthds := methods(u, inter.PkgPath_)

  func methods(u *abi.UncommonType, from string) []abi.Method {
      if u.PkgPath_ == from {
          return u.Methods()          // All methods if same package
      }
      return u.ExportedMethods()      // Only exported if different package
  }
```

  This optimization attempted to improve itab filling performance by filtering out private methods when the interface and type are from different packages.

  However, this optimization relies on interfaceType.PkgPath being set to the interface definition package. Currently, interfaceType.PkgPath is set to the
  package where the interface is used (where the interface conversion happens) rather than the package where the interface is defined. This causes private
  methods to be incorrectly filtered out, leaving itab slots as NULL and causing segmentation faults.

  Solution

  Remove the pre-filtering optimization entirely. Since method names (Name_) already contain package path information:
  - Exported methods: "MethodName"
  - Unexported methods: "package/path.methodName"

  When Name_ matches, the method is guaranteed to be the correct one. The additional package checks are redundant, and comparing all methods is safe without
  risk of adding unexpected functions.

  This PR removes the methods() filtering function and directly uses u.Methods() for matching, bypassing the interfaceType.PkgPath issue entirely.

  Impact

  ✅ Fixes segfaults for:
  - Standard library: go/types.Object, go/ast.Expr, etc.
  - User-defined interfaces with private methods
  - Cross-package interface conversions returning concrete type pointers

  All test cases in the issue now pass.

### DIFF
--- a/.github/actions/setup-deps/action.yml
+++ b/.github/actions/setup-deps/action.yml
@@ -16,6 +16,8 @@ runs:
     - name: Install macOS dependencies
       if: runner.os == 'macOS'
       shell: bash
+      env:
+        HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
       run: |
         brew update
         
--- a/_demo/go/gotoken/main.go
+++ b/_demo/go/gotoken/main.go
@@ -1,0 +1,356 @@
+package main
+
+import (
+	"fmt"
+	"go/token"
+)
+
+func main() {
+	testPos()
+	testToken()
+	testFileSet()
+	testFile()
+	testPosition()
+	testTokenPrecedence()
+	testTokenKeywords()
+	testUtilityFunctions()
+}
+
+func testPos() {
+	fmt.Println("=== Test Pos ===")
+	
+	pos1 := token.Pos(100)
+	pos2 := token.Pos(200)
+	
+	if pos1 != 100 {
+		panic(fmt.Sprintf("Expected pos1 to be 100, got %d", pos1))
+	}
+	if pos2 != 200 {
+		panic(fmt.Sprintf("Expected pos2 to be 200, got %d", pos2))
+	}
+	fmt.Printf("Pos1: %d, Pos2: %d\n", pos1, pos2)
+	
+	if !pos1.IsValid() {
+		panic("Expected pos1.IsValid() to be true")
+	}
+	fmt.Printf("Pos1.IsValid(): %v\n", pos1.IsValid())
+	
+	noPos := token.NoPos
+	if noPos != 0 {
+		panic(fmt.Sprintf("Expected NoPos to be 0, got %d", noPos))
+	}
+	if noPos.IsValid() {
+		panic("Expected NoPos.IsValid() to be false")
+	}
+	fmt.Printf("NoPos: %d, IsValid: %v\n", noPos, noPos.IsValid())
+	
+	fmt.Println("SUCCESS: Pos operations work correctly\n")
+}
+
+func testToken() {
+	fmt.Println("\n=== Test Token Types ===")
+	
+	expectedStrings := map[token.Token]string{
+		token.ADD:    "+",
+		token.SUB:    "-",
+		token.MUL:    "*",
+		token.QUO:    "/",
+		token.LPAREN: "(",
+		token.RPAREN: ")",
+		token.EQL:    "==",
+		token.NEQ:    "!=",
+	}
+	
+	for tok, expected := range expectedStrings {
+		if tok.String() != expected {
+			panic(fmt.Sprintf("Expected %v.String() to be %q, got %q", tok, expected, tok.String()))
+		}
+	}
+	
+	tokens := []token.Token{
+		token.ILLEGAL,
+		token.EOF,
+		token.COMMENT,
+		token.IDENT,
+		token.INT,
+		token.FLOAT,
+		token.IMAG,
+		token.CHAR,
+		token.STRING,
+		token.ADD,
+		token.SUB,
+		token.MUL,
+		token.QUO,
+		token.REM,
+		token.AND,
+		token.OR,
+		token.XOR,
+		token.SHL,
+		token.SHR,
+		token.AND_NOT,
+		token.ADD_ASSIGN,
+		token.SUB_ASSIGN,
+		token.MUL_ASSIGN,
+		token.QUO_ASSIGN,
+		token.REM_ASSIGN,
+		token.AND_ASSIGN,
+		token.OR_ASSIGN,
+		token.XOR_ASSIGN,
+		token.SHL_ASSIGN,
+		token.SHR_ASSIGN,
+		token.AND_NOT_ASSIGN,
+		token.LAND,
+		token.LOR,
+		token.ARROW,
+		token.INC,
+		token.DEC,
+		token.EQL,
+		token.LSS,
+		token.GTR,
+		token.ASSIGN,
+		token.NOT,
+		token.NEQ,
+		token.LEQ,
+		token.GEQ,
+		token.DEFINE,
+		token.ELLIPSIS,
+		token.LPAREN,
+		token.LBRACK,
+		token.LBRACE,
+		token.COMMA,
+		token.PERIOD,
+		token.RPAREN,
+		token.RBRACK,
+		token.RBRACE,
+		token.SEMICOLON,
+		token.COLON,
+	}
+	
+	for _, tok := range tokens {
+		fmt.Printf("Token: %s (String: %q)\n", tok, tok.String())
+	}
+	
+	fmt.Println("SUCCESS: Token types work correctly\n")
+}
+
+func testTokenKeywords() {
+	fmt.Println("\n=== Test Keywords ===")
+	
+	keywords := []token.Token{
+		token.BREAK,
+		token.CASE,
+		token.CHAN,
+		token.CONST,
+		token.CONTINUE,
+		token.DEFAULT,
+		token.DEFER,
+		token.ELSE,
+		token.FALLTHROUGH,
+		token.FOR,
+		token.FUNC,
+		token.GO,
+		token.GOTO,
+		token.IF,
+		token.IMPORT,
+		token.INTERFACE,
+		token.MAP,
+		token.PACKAGE,
+		token.RANGE,
+		token.RETURN,
+		token.SELECT,
+		token.STRUCT,
+		token.SWITCH,
+		token.TYPE,
+		token.VAR,
+	}
+	
+	for _, kw := range keywords {
+		if !kw.IsKeyword() {
+			panic(fmt.Sprintf("Expected %s to be a keyword", kw))
+		}
+		fmt.Printf("Keyword: %s, IsKeyword: %v\n", kw, kw.IsKeyword())
+	}
+	
+	if token.ADD.IsKeyword() {
+		panic("Expected ADD operator to not be a keyword")
+	}
+	if token.IDENT.IsKeyword() {
+		panic("Expected IDENT token to not be a keyword")
+	}
+	
+	fmt.Println("SUCCESS: Keyword checks work correctly\n")
+}
+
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: runtime/internal/runtime/z_face.go [hunk 1]
```diff
@@ -386,7 +386,7 @@ func findMethod(mthds []abi.Method, im abi.Imethod) abi.Text {
 	for _, m := range mthds {
 		mName := m.Name_
 		if mName >= imName {
-			if mName == imName && m.Mtyp_ == im.Typ_ {
+			if mName == imName && m.Mtyp_ == im.Typ_ && (m.Exported() || im.PkgPath() == m.PkgPath()) {
```
visualfc: Name_ 本身格式为 Name 或者是 pkg.Name ( 不导出时 ）所以如果 Name_ 相同，最后两个判断也一定是相同的。

luoliwoshang: 明白，那么基于当前的匹配逻辑来说，把所有abi.Type对应的Method都进入对比是安全的，并不会额外添加一些不预期的函数。当前的 `mthds := methods(u, inter.PkgPath_)`  是提高itab 填充func数组效率的一个方式，但是由于当前interfaceType.PkgPath 设定为了使用包，而非定义包，导致这个优化过程把应该保留的私有方法过滤了，当前Pr将移除当前提前过滤private methods的过程。

luoliwoshang:
@xushiwei

PR #1414, opened 2025-11-20
Fix: Correct initialization order for overlay packages

### LINKED ISSUE #1389: `go/importer` got segement fault when use `importer.ForCompiler`
```go
package main

import (
	"go/importer"
	"go/token"
)

func main() {
	fset := token.NewFileSet()
	imp := importer.ForCompiler(fset, "gc", nil)
	_ = imp
}
```
got 

```bash
llgo/_demo/a on  main [$?] via 🐹 v1.24.6 took 15s 
❯ llgo run main.go    
signal: segmentation fault
```

luoliwoshang:
Fixes #1389

## Problem Summary

When importing `go/importer` package (which depends on `go/build` overlay package), the program crashes with a segmentation fault at runtime. The crash occurs in `io/fs.init$after` when initializing the `io/fs.DirEntry` interface type metadata.

**Crash Stack:**
```
* frame #0: Type.String() + 12               (nil pointer dereference at offset 0x14)
  frame #1: interfaceStr() + 716
  frame #2: Interface() + 236
  frame #3: io/fs.init$after + 23396         (← crash here)
  frame #4: io/fs.init + 44
  frame #5: os.init + 48
  frame #6: fmt.init + 72
  frame #7: go/build.init$hasPatch + 80
  frame #8: go/build.init + 48
```

## Root Cause Analysis

The issue arises from incorrect initialization order interacting with `abiInitType`'s lazy loading mechanism.

### Core Problem: Incorrect Initialization Order

**Problem:** Overlay package's `init$after` executes **before** `init$hasPatch`

**Before Fix:**
```llvm
go/build.init():
  1. call init$after()         ← Executes first (wrong!)
  2. call init$hasPatch()       ← Then initializes dependencies
     └─> call fmt.init()
         └─> call os.init()
             └─> call io/fs.init()
                 └─> call io/fs.init$after()
```

**Consequence:**
- `go/build.init$after` uses `io/fs.FileMode` type
- This triggers premature initialization of `FileMode` (creates type shell only)
- The type is marked as "already processed" by `abiInitType`'s lazy loading mechanism
- Later when `io/fs.init$after` runs, `FileMode` is already non-null

### Interaction with abiInitType Lazy Loading

`abiInitType` uses lazy loading to optimize type initialization:
- Types are only initialized on first use
- Once a type is created, it's marked as "processed"
- Subsequent initialization attempts are skipped

**How This Causes the Bug:**

```
1. go/build.init$after executes first (initialization order error)
   └─> Uses io/fs.FileMode
       └─> FileMode type shell created and marked as "processed"

2. go/build.init$hasPatch executes
   └─> Calls io/fs.init() → io/fs.init$after
       └─> Needs to initialize io/fs.DirEntry interface
           └─> Processes Type() method signature: func() FileMode

3. abiInitType checks FileMode
   ├─> FileMode already exists (marked as "processed")
   ├─> Skips full type initialization (including method types)
   └─> Result: func() FileMode is never created

4. Runtime crash
   └─> DirEntry.Type() method's Imethod.Typ_ = nil
       └─> Accessing nil pointer causes segmentation fault
```

**Key Point:** The lazy loading behavior is correct by design. The problem is that incorrect initialization order causes types to be marked as "processed" before their complete initialization (including method types) can occur.

## Solution

This PR fixes the initialization order problem by ensuring dependencies are initialized before type metadata creation.

### Key Changes

Split `init$hasPatch` into two parts to ensure correct initialization order:

**After Fix:**
```llvm
go/build.init():
  1. call init$hasPatch$deps()  ← Dependency initialization only
     └─> call fmt.init()
         └─> call os.init()
             └─> call io/fs.init()
                 └─> call io/fs.init$after()

  2. call init$after()          ← Type metadata initialization

  3. call init$hasPatch()       ← Global variable initialization
```

**Benefits:**
- ✅ All dependency packages initialized before type metadata creation
- ✅ Types not marked as "used" before `init$after` runs
- ✅ Type metadata created before global variables use those types
- ✅ Prevents premature type shell creation

### Implementation Details

**1. Core Fix (d6d4c60)**
- Generate `init$hasPatch$deps` function to separate dependency initialization
- Implement `findFirstNonInitInstruction()` to split dependency init calls from variable initialization
- Modify overlay init to call in correct order: deps → init$after → hasPatch
- Update `notInit()` to skip `$deps` calls for proper `init$after` placement

**2. Regression Tests (1444f2c)**
- Add `_demo/go/gobuild-1389` test case
- Ensure fix prevents future regressions

**3. Code Refactoring (7ca8ed7)**
- Extract `createInitDepsFunction()` method for better code organization
- Remove unnecessary `hasPatchDepsFunc` field from context
- Simplify `compileBlock()` logic by combining function creation and body generation

## Testing

- ✅ Original issue reproduction case (`go/importer.ForCompiler`)
- ✅ Existing test suite passes
- ✅ New regression tests added

## Notes

**Why This Fix Works:**

By correcting the initialization order, we ensure that:
- All dependency packages complete their initialization (including type metadata)
- Types are fully initialized before being marked as "processed"
- The lazy loading mechanism operates correctly with complete type information

This prevents the scenario where a type is prematurely marked as processed before its method types are created.

### DIFF
--- a/cl/compile.go
+++ b/cl/compile.go
@@ -139,8 +139,6 @@ type context struct {
 	cgoRet     llssa.Expr
 	cgoSymbols []string
 	rewrites   map[string]string
-
-	hasPatchDepsFunc llssa.Function // init$hasPatch$deps function
 }
 
 func (p *context) rewriteValue(name string) (string, bool) {
@@ -478,24 +476,9 @@ func (p *context) compileBlock(b llssa.Builder, block *ssa.BasicBlock, n int, do
 	}
 
 	// Handle init$hasPatch: split dependency init calls into separate function
-	var depsInstrs []ssa.Instruction
 	if doModInit && p.state == pkgHasPatch {
-		// Always create deps function for init$hasPatch (even if empty)
-		// This is needed because overlay init will call it
-		if p.hasPatchDepsFunc == nil {
-			// Remove $hasPatch suffix from function name to get base name
-			fnName := p.fn.Name()
-			baseName := strings.TrimSuffix(fnName, "$hasPatch")
-			depsName := initFnNameOfHasPatchDeps(baseName)
-			p.hasPatchDepsFunc = pkg.NewFunc(depsName, llssa.NoArgsNoRet, llssa.InC)
-		}
-
-		firstNonInit := findFirstNonInitInstruction(instrs)
-		if firstNonInit > 0 {
-			// Split dependency init calls from global var initialization
-			depsInstrs = instrs[:firstNonInit]
-			instrs = instrs[firstNonInit:]
-		}
+		baseName := strings.TrimSuffix(p.fn.Name(), "$hasPatch")
+		instrs = p.createInitDepsFunction(pkg, baseName, instrs)
 	}
 
 	if doModInit {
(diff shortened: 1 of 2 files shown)


visualfc (approved):

luoliwoshang:
@xushiwei

PR #1416, opened 2025-11-24
runtime: implement weak

visualfc:
- runtime/internal/runtime: func `AddCleanupPtr(ptr unsafe.Pointer, cleanup func()) (cancel func())`
- runtime/internal/lib/weak: implement


visualfc:
@xgopilot  code review

Issue #1427, opened 2025-12-01
emb:  esp32c3 print support


luoliwoshang:
- [x] correct libc start https://github.com/goplus/llgo/pull/1435
- [x] c.Printf https://github.com/goplus/llgo/pull/1440
* println/runtime
   - [x] illgeal atomic direction https://github.com/goplus/llgo/pull/1443 
- [ ] other error at println 

---

```go
package main

func main() {
	println("yes")
}
```

got follow
```bash 
llgo/_demo/a on  main [$!?] via 🐹 v1.24.10 
❯ llgo run -target esp32c3 .
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
Generating firmware image: /var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/a-3674589995.elf -> /var/folders/5j/sgtxqmbn1hbdqtgx5kkp6y700000gn/T/a-3674589995.bin (format: esp32c3, detail: )
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
esptool.py v4.9.0
Serial port /dev/cu.usbmodem1101
Connecting...
Chip is ESP32-C3 (QFN32) (revision v0.4)
Features: WiFi, BLE, Embedded Flash 4MB (XMC)
Crystal is 40MHz
USB mode: USB-Serial/JTAG
MAC: dc:b4:d9:9a:ce:04
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Flash will be erased from 0x00000000 to 0x0000dfff...
Compressed 57296 bytes to 30471...
Wrote 57296 bytes (30471 compressed) at 0x00000000 in 0.4 seconds (effective 1200.5 kbit/s)...
Hash of data verified.

Leaving...
Hard resetting via RTS pin...
[...]

PR #1440, opened 2025-12-05
emb:esp32c3:implement _write by jtag output

luoliwoshang:
part of  https://github.com/goplus/llgo/issues/1427

# ESP32-C3 USB Serial JTAG Output Support

## Overview

Implements USB Serial JTAG output support for ESP32-C3, enabling `printf()` and other standard output functions to work in bare-metal applications without external UART connections.

## Why USB Serial/JTAG Instead of UART?

**Core reason: All ESP32-C3 chips have built-in USB Serial/JTAG, but not all development boards have USB-UART converter chips!**

According to [ESP-IDF USB Serial/JTAG Controller Documentation](https://docs.espressif.com/projects/esp-idf/en/v5.5.1/esp32c3/api-guides/usb-serial-jtag-console.html):

> "Generally, ESP chips implement a serial port using UART and can be connected to a serial console emulator on a host/PC via an **external USB-UART bridge chip**. However, on ESP chips that contain a USB Serial/JTAG Controller, the CDC-ACM portion of the controller implements a serial port that is **connected directly to a host/PC**, thus **does not require an external USB-UART bridge chip**."

Many low-cost development boards (e.g., XIAO ESP32C3, SuperMini) do not include external USB-UART chips (CH340/CP2102) to save costs. Their USB ports connect directly to the chip's built-in USB Serial/JTAG (GPIO18/19). If UART output were used, these boards would have no serial output at all.

TinyGo encountered the same issue in [Issue #3631](https://github.com/tinygo-org/tinygo/issues/3631) and resolved it via [PR #4011](https://github.com/tinygo-org/tinygo/pull/4011) by changing the default serial from UART to USB Serial/JTAG:

```diff
// targets/esp32c3.json
- "serial": "uart",
+ "serial": "usb",
```

Source: [TinyGo commit d7c77b67](https://github.com/tinygo-org/tinygo/commit/d7c77b6761f25e3ded9863503ca37dc1f870ac0e)

## Key Features

- ✅ **USB Serial JTAG Output** - Implemented via `_write()` system call
- ✅ **Bare-Metal Implementation** - No FreeRTOS or driver layer required
- ✅ **Non-Blocking** - Discards characters when FIFO is full, never hangs
- ✅ **Automatic CRLF Conversion** - `\n` automatically converted to `\r\n`
- ✅ **Direct Register Access** - Based on ESP-IDF HAL layer, but without full framework

## Implementation Sources

This implementation is based on official code from **ESP-IDF v6.0**, streamlined and adapted:

### 1. USB Serial JTAG Register Structure

**Source**: [`components/soc/esp32c3/register/soc/usb_serial_jtag_struct.h`](https://github.com/espressif/esp-idf/blob/release/v6.0/components/soc/esp32c3/register/soc/usb_serial_jtag_struct.h)

Defines the register layout for USB Serial JTAG peripheral, including:
- `ep1` - TX/RX FIFO data register
- `ep1_conf` - FIFO status and control register
- Other control and status registers

Mapped to hardware address `0x60043000`.

### 2. HAL Layer Low-Level Functions

**Source**: [`hal/esp32c3/include/hal/usb_serial_jtag_ll.h`](https://github.com/espressif/esp-idf/blob/release/v6.0/components/hal/esp32c3/include/hal/usb_serial_jtag_ll.h)

Implements three key inline functions:
- `usb_serial_jtag_ll_txfifo_flush()` - Flush TX FIFO
- `usb_serial_jtag_ll_txfifo_writable()` - Check if TX FIFO has space
- `usb_serial_jtag_ll_write_txfifo()` - Write data to TX FIFO

### 3. System Call Implementation

**Source**: Referenced from [`esp_driver_usb_serial_jtag/src/usb_serial_jtag_vfs.c`](https://github.com/espressif/esp-idf/blob/release/v6.0/components/esp_driver_usb_serial_jtag/src/usb_serial_jtag_vfs.c)

Implements `_write()` system call to redirect standard output to USB Serial JTAG. Main features:
- Supports stdout and stderr
- Non-blocking write (discards on FIFO full)
- Automatic LF → CRLF conversion

### 1. Memory Layout (`targets/esp32c3.memory.ld`)

Added USB Serial JTAG peripheral base address:

```ld
PROVIDE ( USB_SERIAL_JTAG = 0x60043000 );
```

### 2. Device Implementation (`targets/device/esp/esp32c3.c`)

Complete USB Serial JTAG driver implementation (~400 lines), including:
- Structure definitions (from ESP-IDF SOC layer)
- HAL layer functions (from ESP-IDF HAL layer)
- `_write()` system call (referenced from ESP-IDF VFS layer)

## Dependencies

**Required**: https://github.com/goplus/newlib/pull/9

Newlib PR #9 declares `_write()` as a **weak symbol**, allowing us to provide a custom implementation:

```c
// libgloss/riscv/esp/syscalls.c
__attribute__((weak))
ssize_t _write(int file, const char *ptr, size_t len) {
    // Default UART implementation (can be overridden)
}
```

During linking:
```
esp32c3.o: _write (strong symbol)  ← Our JTAG implementation (wins)
syscalls.o: _write (weak symbol)   ← Newlib's UART implementation (ignored)
```

## Usage Example

```go
package main

import "github.com/goplus/lib/c"

func main() {
    c.Printf(c.Str("Hello from ESP32-C3 via USB Serial JTAG!\n"))
}
```

**Build and run** (automatic flash + monitor):
```bash
llgo run -a -target esp32c3 .
```

**Expected output** (displayed directly via USB cable):
```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x7 (TG0WDT_SYS_RST),boot:0xd (SPI_FAST_FLASH_BOOT)
Saved PC:0x40052b02
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fc855e0,len:0xfc
load:0x3fc856dc,len:0x4
load:0x3fc856e0,len:0x44
load:0x40380000,len:0x1578
load:0x40381578,len:0x68
entry 0x40380000
Hello from ESP32-C3 via USB Serial JTAG!
```

### Why Not Use ESP-IDF Directly?

1. **Code Size**: ESP-IDF framework is large, containing many unnecessary features
2. **Complex Dependencies**: Requires FreeRTOS, driver layer, VFS layer, etc.
3. **Startup Overhead**: Long initialization time, unsuitable for simple applications
4. **Bare-Metal Target**: llgo's embedded targets prioritize lightweight and direct control

### Our Approach

✅ **Extract Core**: Only extract register definitions and key functions
✅ **Simplified Implementation**: Remove VFS layer, interrupts, DMA, and other complex features
✅ **Maintain Compatibility**: Register operations fully compatible with ESP-IDF
✅ **Document Sources**: Code clearly annotated with sources for easy maintenance and updates

### ESP-IDF Official Documentation
- [USB Serial JTAG Controller](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-guides/usb-serial-jtag-console.html)
- [ESP32-C3 Technical Reference Manual](https://documentation.espressif.com/esp32-c3_technical_reference_manual_cn.html)

### ESP-IDF Source Code References
- [SOC Layer Register Definitions](https://github.com/espressif/esp-idf/blob/release/v6.0/components/soc/esp32c3/register/soc/usb_serial_jtag_struct.h)
- [HAL Layer LL Functions](https://github.com/espressif/esp-idf/blob/release/v6.0/components/hal/esp32c3/include/hal/usb_serial_jtag_ll.h)
- [VFS Layer Implementation](https://github.com/espressif/esp-idf/blob/release/v6.0/components/esp_driver_usb_serial_jtag/src/usb_serial_jtag_vfs.c)

### Dependency PR
- [goplus/newlib#9](https://github.com/goplus/newlib/pull/9) - Weak symbol support for `_write()`

---

## Commit History

This PR includes the following commits:

1. **emb:esp32c3:USB_SERIAL_JTAG load** - Add peripheral base address
2. **emb:esp32c3:implement _write by jtag output** - Implement complete driver
3. **build: upgrade newlib-esp32 from patch3 to patch4** - Upgrade Newlib with weak symbol support
4. **emb:esp32c3:add source reference comments** - Add source annotations

---

**Fix Type**: Feature
**Priority**: Medium (improves ESP32-C3 development experience)
**Backward Compatible**: Yes (does not affect existing code)

---

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/targets/device/esp/esp32c3.c
+++ b/targets/device/esp/esp32c3.c
@@ -347,23 +347,23 @@ static inline int usb_serial_jtag_ll_write_txfifo(const uint8_t *buf, uint32_t w
 #include <sys/types.h>
 
 /**
- * @brief Write one character to USB Serial JTAG (non-blocking, simplified)
+ * @brief Write one character to USB Serial JTAG (blocking)
  *
- * If FIFO is full, character is dropped silently.
- * This is a bare-metal implementation without timeout or retry logic.
+ * Waits until FIFO has space before writing.
  */
 static void usb_serial_jtag_write_char(char c) {
-    // Check if TX FIFO has space
-    if (usb_serial_jtag_ll_txfifo_writable()) {
-        uint8_t byte = (uint8_t)c;
-        usb_serial_jtag_ll_write_txfifo(&byte, 1);
-
-        // Flush on newline to ensure output is displayed immediately
-        if (c == '\n') {
-            usb_serial_jtag_ll_txfifo_flush();
-        }
+    // Wait until TX FIFO has space (blocking)
+    while (!usb_serial_jtag_ll_txfifo_writable()) {
+        // Busy wait
     }
-    // If FIFO is full, character is dropped (non-blocking behavior)
+
+    uint8_t byte = (uint8_t)c;
+    usb_serial_jtag_ll_write_txfifo(&byte, 1);
+
+    // Flush on newline to ensure output is displayed immediately
+    if (c == '\n') {
+        usb_serial_jtag_ll_txfifo_flush();
+    }
 }
 
 /**


cpunion:
- _write is non-blocking and drops on full. The USB Serial JTAG TX FIFO is only 64B and the UART TX FIFO is ~128B on ESP32-C3, so long/bursty prints will be lost when those buffers fill.
- Make the console sink pluggable (JTAG / UART / noop / custom). Allow overriding from Go so users can route output elsewhere (e.g., LCD) instead of being hardwired to JTAG.

luoliwoshang:
> Make the console sink pluggable (JTAG / UART / noop / custom). Allow overriding from Go so users can route output elsewhere (e.g., LCD) instead of being hardwired to JTAG.

If the _write implementation in the current `targets/device/esp/esp32c3.c` is made a weak symbol, under the current linking rules, it will actually use the _write symbol from newlib. By "overriding," do you mean implementing it as a weak symbol?

targets/device/esp/esp32c3.c
```diff
@@ -0,0 +1,401 @@
...
+
+/**
+ * @brief Write one character to USB Serial JTAG (non-blocking, simplified)
+ *
+ * If FIFO is full, character is dropped silently.
+ * This is a bare-metal implementation without timeout or retry logic.
+ */
+static void usb_serial_jtag_write_char(char c) {
+    // Check if TX FIFO has space
+    if (usb_serial_jtag_ll_txfifo_writable()) {
```
MeteorsLiu: It seems it's not a good implement, when the JTAG buffer is full, we should wait until available, that's we called blocked `write`

MeteorsLiu: However, in the future, we may support non-blocking way to implement it. See #1445

luoliwoshang: > _write is non-blocking and drops on full. The USB Serial JTAG TX FIFO is only 64B and the UART TX FIFO is ~128B on ESP32-C3, so long/bursty prints will be lost when those buffers fill.

@cpunion 

In the ESP-IDF implementation, it will wait 50 milliseconds before choosing to discard the data. Perhaps here, to ensure data is not lost, we could keep blocking and waiting for usb_serial_jtag_ll_txfifo_writable()?

```c
static void usb_serial_jtag_tx_char_no_driver(int fd, int c)
{
    uint8_t cc = (uint8_t)c;
    // Try to write to the buffer as long as we still expect the buffer to have
    // a chance of being emptied by an active host. Just drop the data if there's
    // no chance anymore.
    // When we first try to send a character and the buffer is not accessible yet,
    // we wait until the time has been more than TX_FLUSH_TIMEOUT_US since we successfully
    // sent the last byte. If it takes longer than TX_FLUSH_TIMEOUT_US, we drop every
    // byte until the buffer can be accessible again.
    do {
        if (usb_serial_jtag_ll_txfifo_writable()) {
            usb_serial_jtag_ll_write_txfifo(&cc, 1);
            if (c == '\n') {
                //Make sure line doesn't linger in fifo
                usb_serial_jtag_ll_txfifo_flush();
            }
            //update time of last successful tx to now.
            s_ctx.last_tx_ts = esp_timer_get_time();
            break;
        }
    } while ((esp_timer_get_time() - s_ctx.last_tx_ts) < TX_FLUSH_TIMEOUT_US);

}

```

luoliwoshang: current implement as blocking https://github.com/goplus/llgo/pull/1440/commits/b45b56c9630bbf5246eb860331c9491eb3297eaf

luoliwoshang:
@xushiwei

PR #1450, opened 2025-12-11
ssa: abi type global constant data

visualfc:
- [x] types.Basic
- [x] types.Pointer
- [x] types.Slice
- [x] types.Array
- [x] types.Chan
- [x] types.Map
- [x] types.Struct / Closure
- [x] types.Signature
- [x] types.Struct hash
- [x] types.Array hash
- [x] types.Interface
- [x] types.Named
- [x] UncommonType
- [x] Methods
- [x] abi types init for reflect

```
struct Type {
	Common ( _type )
	Extended ( optional )
}

struct {
	Type
	UncommonType
	[N]Method
}
```

const global data
```
@_llgo_int = weak_odr constant %"github.com/goplus/llgo/runtime/abi.Type" { i64 8, i64 0, i32 -25294021, i8 8, i8 8, i8 8, i8 34, { ptr, ptr } { ptr @"__llgo_stub.github.com/goplus/llgo/runtime/internal/runtime.memequal64", ptr null }, ptr null, %"github.com/goplus/llgo/runtime/internal/runtime.String" { ptr @3, i64 3 }, ptr @"*_llgo_int" }, align 8
@3 = private unnamed_addr constant [3 x i8] c"int", align 1
@"*_llgo_int" = weak_odr constant %"github.com/goplus/llgo/runtime/abi.PtrType" { %"github.com/goplus/llgo/runtime/abi.Type" { i64 8, i64 8, i32 -939606833, i8 10, i8 8, i8 8, i8 54, { ptr, ptr } { ptr @"__llgo_stub.github.com/goplus/llgo/runtime/internal/runtime.memequalptr", ptr null }, ptr null, %"github.com/goplus/llgo/runtime/internal/runtime.String" { ptr @3, i64 3 }, ptr null }, ptr @_llgo_int }, align 8
```

binary size optimization

- no abi types
```
package main

func main() {
	println("hello world")
}
```
```
-rwxr-xr-x  1 apple  staff    63920 Dec 25 12:22 hello1_new
-rwxr-xr-x  1 apple  staff   189320 Dec 25 12:22 hello1_old
-rwxr-xr-x  1 apple  staff    77096 Dec 25 12:36 hello1_tinygo
-rwxr-xr-x  1 apple  staff  1553216 Dec 25 12:31 hello1_go124
```

- use reflect. no need abi types init.
```
package main

import (
	"fmt"
)

func main() {
	fmt.Println("hello world")
}
```
```
-rwxr-xr-x  1 apple  staff  1724424 Dec 25 12:11 hello2_new
-rwxr-xr-x  1 apple  staff  2371856 Dec 25 12:12 hello2_old
-rwxr-xr-x  1 apple  staff   270688 Dec 25 12:37 hello2_tinygo
-rwxr-xr-x  1 apple  staff  2358096 Dec 25 12:32 hello2_go124
```

- use reflect. need abi types init. ( reflect.SliceOf )
```
package main

import (
	"fmt"
	"reflect"
)

func main() {
	fmt.Println("hello world")
	t := reflect.SliceOf(reflect.TypeOf(0))
	fmt.Println(t == reflect.TypeOf([]int{}))
}
```

```
-rwxr-xr-x  1 apple  staff  2262472 Dec 25 17:16 hello3_new
-rwxr-xr-x  1 apple  staff  2381104 Dec 25 12:15 hello3_old
-rwxr-xr-x  1 apple  staff   271672 Dec 25 12:38 hell3_tinygo ( panic: unimplemented: reflect.SliceOf() )
-rwxr-xr-x  1 apple  staff  2371072 Dec 25 12:33 hello3_go124
```

hello3 demo: need abi types init
```
@"init$abitypes$array" = constant [1781 x ptr] [ptr @"*[0]*_llgo_any", ptr @"*[0]_llgo_uint8",  ... , ptr @"map[_llgo_uint64]_llgo_bool"], align 8
@"init$abitypes$slice" = constant %"github.com/goplus/llgo/runtime/internal/runtime.Slice" { ptr @"init$abitypes$array", i64 1781, i64 1781 }, align 8

define void @"init$abitypes"() {
_llgo_0:
  call void @"github.com/goplus/llgo/runtime/internal/runtime.initTypes"(ptr @"init$abitypes$slice")
  ret void
}

define i32 @main(i32 %0, ptr %1) {
_llgo_0:
  store i32 %0, ptr @__llgo_argc, align 4
  store ptr %1, ptr @__llgo_argv, align 8
  call void @"github.com/goplus/llgo/runtime/internal/runtime.init"()
  call void @"init$abitypes"()
  call void @runtime.init()
  call void @command-line-arguments.init()
  call void @command-line-arguments.main()
  ret i32 0
}
```


visualfc:
@xgopilot code review

cpunion:
Hi, I ran into two blockers while expanding the ABI coverage tests.

1. **GC metadata is still missing.** All `_llgo_*` constants emitted for the new `cl/_testrt/abi_gcdata` case have `GCData = null`, so pointer fields are treated as pointer-free. The new test below (now in `test/abi/abi_test.go`) passes with the stock Go toolchain but fails with llgo:

```
$ go run ./cmd/llgo test ./test/abi
--- FAIL: TestGCKeepsChildren (0.04s)
    abi_test.go:81: child never finalized after parent dropped
```

Because the GC never sees the child pointer, its finalizer never runs. We need either real bitmaps or at least `TFlagGCMaskOnDemand` wired up before these ABI tables can be trusted. For reference, `runtime/internal/lib/runtime/mfinal.go` still contains a `panic("todo: runtime.SetFinalizer")`, so even the simple finalizer test in `test/finalizer_test.go` fails under llgo (`finalizer did not run`). That probably needs fixing first.

2. **`NeedAbiInit` fires even when there is nothing to register.** The new `cl/_testrt/abi_reflectinit` program only calls `reflect.SliceOf`, yet `prog.abiSymbol` stays empty and the generated `out.ll` contains no `init$abitypes`. The linker still emits an `init$abitypes` stub solely because touching `reflect.SliceOf` sets `NeedAbiInit`, so any package that uses those helpers drags in an extra init with an empty slice. For example, the emitted LLVM IR looks like:

```
@"init$abitypes$array" = constant [0 x ptr] [], align 8
@"init$abitypes$slice" = constant %runtime/internal/runtime.Slice { ptr @"init$abitypes$array", i64 0, i64 0 }, align 8

define void @"init$abitypes"() {
_llgo_0:
  call void @"runtime/internal/runtime.initTypes"(%runtime/internal/runtime.Slice @"init$abitypes$slice")
  ret void
}
```

Every `reflect.SliceOf` call pulls in that empty stub even though there is nothing to register. Could we gate `NeedAbiInit` on `len(prog.abiSymbol) > 0` (or split the tables per package) to avoid paying for a useless init?

Happy to push the new tests once we have a plan for GC data generation, finalizer plumbing, and the init heuristic. Thanks!

cpunion:
Adding the new GC regression test (now in `test/abi/abi_test.go`) so it's easy to find:

```go
type node struct {
    child   *node
    padding [2]uintptr
}

func TestGCKeepsChildren(t *testing.T) {
    var finalized atomic.Int32
    makeParent := func() *node {
        child := &node{}
        runtime.SetFinalizer(child, func(*node) {
            finalized.Add(1)
        })
        return &node{child: child}
    }

    parent := makeParent()
    for i := 0; i < 10; i++ {
        runtime.GC()
        runtime.Gosched()
    }
    if finalized.Load() != 0 {
        t.Fatalf("child finalized while parent is still alive: %d", finalized.Load())
    }
    runtime.KeepAlive(parent)

    parent = nil
    for i := 0; i < 200 && finalized.Load() == 0; i++ {
        runtime.GC()
        runtime.Gosched()
    }
    if finalized.Load() == 0 {
        t.Fatalf("child never finalized after parent dropped")
    }
}
```

This passes under the stock Go toolchain but fails with llgo because GCData for the generated `_llgo_*` types is still nil.

PR #1516, opened 2025-12-30
Record compiler hash in build fingerprints

cpunion:
## Summary
- compute a deterministic hash for development compiler builds based on binary metadata
- add LLGO compiler hash into the manifest env so cache entries invalidate when the compiler changes

## Testing
- go test ./...


internal/env/version.go
```diff
@@ -29,12 +35,40 @@ const (
 // set by the linker.
 var buildVersion string
 
+// compilerHash caches the LLGo compiler fingerprint. Release builds may set it
+// at link time, while development builds compute it from the running binary.
+var compilerHash string
+
+// init precomputes the compilerHash for development builds so cache entries can
+// observe compiler updates without hashing the full executable contents.
+func init() {
```
xushiwei: This shows that `compilerHash` is a feature with global side effects (i.e., it will be linked regardless of whether the user proactively uses it). This type of feature might not be suitable for a basic library like `xgo/env`. After all, `compilerHash` doesn't fall within the scope of interest of the standard library; its only anticipated user is currently llgo itself. Therefore, it should probably be placed in `cmd/llgo`.

Issue #1538, opened 2026-01-08
cl/_testrt/cast: error with panic


luoliwoshang:
`goplus/llgo/cl/_testrt/cast` a3db8d0e13c1c3526d517fb54aa39632c02f457f

```bash
cl/_testrt/cast on  via 🐹 v1.24.10 
❯ llgo run .
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
panic: error

[0x0253C8C0 github.com/goplus/llgo/cl/_testrt/cast.cvt32Fto8U+0x31, SP = 0x48]
[0x0253CD30 github.com/goplus/llgo/cl/_testrt/cast.main+0x2b, SP = 0x198]
[0x0253CF78 main+0x4, SP = 0x24]
exit status 2
```

PR #1544, opened 2026-01-12
fix(cl):same float-to-int cast overflow behavior with go

luoliwoshang:
## Summary
- Fix float->int conversions for narrow integer targets by routing through 64-bit conversion and truncation.
- Preserve Go wrap/truncate semantics for out-of-range float values.
- Avoid LLVM poison results from direct fptosi/fptoui on narrow types.

## Fixed
- Fixes https://github.com/goplus/llgo/issues/1538

## Problem
The repro from issue #1538 panics because LLVM's `fptosi`/`fptoui` on narrow integer types yields poison for out-of-range inputs. That poison is later used and results diverge from Go's expected behavior.

## Root Cause
`ssa/expr.go` used `llvm.CreateFPToSI/FPToUI` directly with the destination integer type. For small widths, the conversion is undefined/poison when the input float is outside the representable range.

## Fix
- Introduce a helper that converts float->int via 64-bit first, then truncates to the target width.
- For unsigned targets, split negative vs non-negative paths to match Go's behavior before truncation.
- Keep existing behavior for >=64-bit integer targets (documented inline in `ssa/expr.go`).

## References
- LLVM `fptoui` semantics (poison on overflow): https://llvm.org/docs/LangRef.html#fptoui-to-instruction
- LLVM `fptosi` semantics (poison on overflow): https://llvm.org/docs/LangRef.html#fptosi-to-instruction
- Go spec (numeric conversions): https://go.dev/ref/spec#Conversions

## Files Touched
- ssa/expr.go
- _demo/go/issue1538/main.go
- _demo/go/issue1538-floatcvtuint-over/main.go
- cl/_testrt/cast/out.ll

## Verification
- llgo run _demo/go/issue1538-floatcvtuint-over/main.go

## Notes
- This change is intentionally scoped to float->int conversions; int->int and int->float paths are unchanged.

### DIFF
--- a/ssa/expr.go
+++ b/ssa/expr.go
@@ -936,6 +936,10 @@ func castFloatToInt(b Builder, x llvm.Value, typ Type) llvm.Value {
 	if dstSize < 8 {
 		i64 := b.Prog.Int64()
 		if typ.kind == vkUnsigned {
+			// note(zzy): for unsigned targets, split negative vs non-negative.
+			// Negative values need signed expansion (FPToSI) before truncation;
+			// non-negative values can use unsigned expansion (FPToUI). This avoids
+			// direct float->narrow-unsigned conversions and preserves wrap/trunc behavior.
 			zero := llvm.ConstNull(x.Type())
 			isNeg := b.impl.CreateFCmp(llvm.FloatOLT, x, zero, "")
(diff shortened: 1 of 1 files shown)


luoliwoshang:
@xushiwei

PR #1435, opened 2025-12-03
fix(esp32c3): use newlib startup and merge init_array to rodata for println support

luoliwoshang:
part of https://github.com/goplus/llgo/issues/1438
part of https://github.com/goplus/llgo/issues/1427

### Background

Currently, `targets/riscv.json` contains TinyGo's startup files `start.S` and `handleinterrupt.S`, which affect all RISC-V chips (including ESP32-C3) through the inheritance mechanism.

ESP32-C3 requires newlib's standard startup process (`crt0-riscv32-unknown-none`). This startup file calls `__libc_init_array` to execute C library constructors (such as `board_init()`), but it is currently overridden by TinyGo's `start.S`, resulting in incomplete C library initialization.

#### Core Approach

1. **Create riscv-basic.json**: Pure configuration base class without `extra-files`
2. **TinyGo Path**: `riscv.json` inherits `riscv-basic` and adds startup files (start.S)
3. **LLGo Path**: `riscv-nostart.json` inherits `riscv-basic` (no startup files)
4. **ESP32-C3**: Inherits `riscv32-nostart`, will not include TinyGo's `start.S`

#### Inheritance Relationship Diagram

```mermaid
graph TD
    A[riscv-basic.json<br/>Pure config, no extra-files] --> B[riscv.json<br/>+ start.S<br/>+ handleinterrupt.S]
    A --> C[riscv-nostart.json<br/>no extra-files<br/>no startup files]

    B --> D[riscv32.json<br/>TinyGo 32-bit config]
    C --> E[riscv32-nostart.json<br/>LLGo 32-bit config<br/>no startup files]

    D --> F[Other TinyGo RISC-V chips<br/>auto get start.S]
    E --> G[esp32c3.json<br/>no start.S]
```

##### 1. Create `targets/riscv-basic.json`
Copy all configurations from `riscv.json`, **remove `extra-files` field**:

```json
{
	"goos": "linux",
	"goarch": "arm",
	"build-tags": ["tinygo.riscv", "baremetal", "linux", "arm"],
	"gc": "conservative",
	"linker": "ld.lld",
	"rtlib": "compiler-rt",
	"libc": "picolibc",
	"cflags": [
		"-Werror",
		"-mno-relax",
		"-fno-exceptions", "-fno-unwind-tables", "-fno-asynchronous-unwind-tables",
		"-ffunction-sections", "-fdata-sections"
	],
	"ldflags": [
		"--gc-sections"
	],
[...]

PR #1549, opened 2026-01-13
feat(esp32): enable semihosting exit in QEMU

luoliwoshang:
## Summary
This PR enables reliable **QEMU exit** for the existing `esp32` target (Xtensa) by switching to newlib's **exit-only semihosting** path.

Goal: keep normal ESP32 runtime behavior for non-exit syscalls, but let `llgo run -target=esp32 -emulator` terminate QEMU cleanly for test automation.

## Motivation
Before this change, ESP32 emulator runs could end via an illegal-instruction panic path without a deterministic semihosting exit signal, which made emulator-based comparison tests unstable.

At the same time, enabling full semihosting is not desired here because it can reroute regular I/O behavior. For ESP32 we only need semihosting on `_exit`.

### 1) newlib integration for Xtensa (ESP32)
- Bump LLGO's ESP32 newlib dependency:
  - from `esp-4.3.0_20250211-patch5`
  - to `esp-4.3.0_20250211-patch6`
- Update libc config tests accordingly.

In LLGO's Xtensa newlib build config:
- Compile `libgloss/xtensa/sim-call.S` (semihost call entry support).
- Add `-DQEMU_SEMIHOSTING_EXIT_ONLY` for Xtensa libgloss/syscalls build.
- Include `libgloss/libnosys/_exit.c` in the build.
- Remove `-DGETREENT_PROVIDED` in Xtensa newlib compile flags so newlib's own reent path is used instead of relying on external stubs.

### 2) ESP32 emulator command
- Update `targets/esp32.json` to run QEMU with semihosting enabled:
  - add `-semihosting` to `qemu-system-xtensa` arguments.

### 3) CI and smoke test improvements
- Replace `test_esp32c3_startup.sh` with a generalized script:
  - `_demo/embed/test-esp-serial-startup.sh`
- Test flow now includes:
  - Common smoke for **ESP32-C3 + ESP32** (build + emulator run)
  - ESP32-C3-specific startup regressions (`__libc_init_array`, `.init_array` in `.rodata`, BIN segment check)
- Stream emulator output in real time during smoke tests (instead of only post-capture).
- Update embed dependency setup to install and verify **both**:
  - `qemu-system-riscv32`
  - `qemu-system-xtensa`

## Newlib Behavior Clarification
With patch6 + `QEMU_SEMIHOSTING_EXIT_ONLY` in Xtensa:
- `_exit` uses semihosting to notify QEMU and terminate emulator runs cleanly.
- Non-exit paths (e.g. regular output path) stay on the existing board/syscall flow.

So this is intentionally **not** full semihosting; it is **exit-only semihosting**.

## Validation
Local checks performed:
- `llgo run -a -target=esp32 -emulator .`
- `bash _demo/embed/test-esp-serial-startup.sh`
- Verified QEMU tool install script provides both Xtensa and RISC-V executables.

If local cache still points to an older newlib bundle, clear and rebuild once:
```bash
rm -rf ~/Library/Caches/llgo/crosscompile/newlib-*
```

## Scope / Compatibility
- No new target introduced; this applies to the existing `esp32` target.
- Change is focused on emulator termination correctness while keeping non-exit behavior aligned with prior ESP32 expectations.


_demo/embed/test-esp-serial-startup.sh
```diff
@@ -0,0 +1,218 @@
...
+run_emulator_smoke "esp32c3-basic" "ESP32-C3" "Hello World"
+
+build_target "esp32" "$ESP32_PREFIX" "ESP32"
+run_emulator_smoke "esp32" "ESP32" "Hello World"
+
+echo ""
+echo "=== ESP32-C3 Specific Regression Tests ==="
+
+echo ""
+echo "=== Regression 1: Verify newlib startup (calls __libc_init_array) ==="
```
MeteorsLiu: This is the regression test script for #1435, it seems that it's unrelated with `esp32`, consider to separate it into a independent script?

luoliwoshang: Yes,It's only for esp32c3's regression,ok,i will keep it at `_demo/embed/test_esp32c3_startup.sh`,and only move the build & run emulator smoke to this public script `_demo/embed/test-esp-serial-startup.sh`

luoliwoshang: Fix at https://github.com/goplus/llgo/pull/1549/commits/fe61012441de1f5d13a5563b578918cad7c5697a

_demo/embed/test_esp32c3_startup.sh
```diff
@@ -232,29 +232,11 @@ else
...
-if [ "$LAST_LINE" = "Hello World" ]; then
-    echo "✓ PASS: QEMU output ends with Hello World"
-else
-    echo "✗ FAIL: QEMU output mismatch"
-    echo "Last line: $LAST_LINE"
-    echo ""
-    echo "Full output:"
-    echo "$RUN_OUT"
-    exit 1
-fi
```
luoliwoshang, line 244: NOTE:verify at [_demo/embed/test-esp-serial-startup.sh](https://github.com/goplus/llgo/pull/1549/files#diff-a0747b0c792ef17533fb12b6ac1954aaa4fdf89c578a4997d3188b04cd7b74ef)

MeteorsLiu (approved):

PR #1568, opened 2026-01-16
Implement register-based closure ctx

cpunion:
## Summary
This PR updates LLGo’s closure ABI and calling convention: use a reserved register for ctx when available; otherwise pass ctx as an implicit first parameter with a conditional call; represent closures as pointers to `funcval` with inline env.

## ABI / Representation
- Closures are represented as `type funcval struct { fn *func; hasCtx uintptr; env ... }` and `type closure = *funcval`; the ABI only sees the pointer.
- `hasCtx` keeps the header size fixed and supports conditional calls on no‑reg targets.
- `env` is inline after the header; plain functions have no env (object size = 2 pointers).
- C function pointers are first‑class: a `funcval` can point directly at a C symbol, so the call site can use the real address without wrapper stubs.

## Calling Convention
- With a ctx register: `write_ctx(env_ptr)` then call `fn(args...)`; `env_ptr = closure_ptr + 2*ptrSize`.
- Without a ctx register: conditionally call `fn(ctx, args...)` vs `fn(args...)` based on `hasCtx`.

## getClosurePtr
- `getClosurePtr` returns `&env[0]` (pointer to the first env slot).
- On ctx‑register targets it reads the ctx register to get the env base.
- On no‑reg targets it uses the explicit ctx parameter as the env base.

## Context Register Mapping
| GOARCH | Register | Notes |
|---|---|---|
| amd64 | mm0 | use `-msse2` to free MMX |
| 386 | mm0 | use `-msse2` to free MMX |
| arm64 | x26 | reserved via clang target-feature |
| riscv64 | x27 | reserved via clang target-feature |
| wasm | - | conditional ctx param |
| arm | - | conditional ctx param |

Native builds reserve the ctx reg via clang target-feature `+reserve-<reg>` (arm64/riscv64).
For caller-saved x86, inline asm uses a memory clobber; callee-saved targets do not.

## Example IR (closure + C func)
Example Go code:
```go
func cfunc(i int64)

func main() {
  var fn func(i int64)
  fn = cfunc
  fn(0)

  var i int64 = 0
  fn = func(v int64) { i = v }
  fn(0)
}
```

### With ctx register (arm64/riscv64/x86*)
Caller (`main`) writes ctx register and calls the real function symbol:
```llvm
; fn = cfunc
store ptr @__llgo_closure_const$cfunc, ptr %fn_slot
; fn(0)
%fv = load ptr, ptr %fn_slot
%fnptr = load ptr, ptr %fv
%env_base = getelementptr i8, ptr %fv, i64 16  ; &env[0]
call void @llvm.write_register.i64(metadata !"x26", i64 %env_base)
call void %fnptr(i64 0)

; fn = closure
%fv2 = call ptr @AllocU(i64 24)          ; {fn,hasCtx,env0}
store ptr @main$1, ptr %fv2
store i64 1, ptr (getelementptr i8, ptr %fv2, i64 8)
store ptr %i, ptr (getelementptr i8, ptr %fv2, i64 16)
store ptr %fv2, ptr %fn_slot
; fn(0) ... same pattern ...
```
Closure body (`main$1`) reads ctx register at entry:
```llvm
define void @main$1(i64 %v) {
entry:
  %env_base = call i64 @llvm.read_register.i64(metadata !"x26")
  %env0p = inttoptr i64 %env_base to ptr
  %i_ptr = load ptr, ptr %env0p
  store i64 %v, ptr %i_ptr
  ret void
}
```
C function remains a normal symbol:
```llvm
declare void @cfunc(i64)
```

### Without ctx register (wasm/arm)
Caller (`main`) branches on `hasCtx` to pick the correct signature:
```llvm
%has = load i64, ptr (getelementptr i8, ptr %fv, i64 8)
%has_i1 = icmp ne i64 %has, 0
br i1 %has_i1, label %with, label %plain

with:
  %fnptr = load ptr, ptr %fv
  %env_base = getelementptr i8, ptr %fv, i64 16
  %fp1 = bitcast ptr %fnptr to ptr (ptr, i64)*
  call void %fp1(ptr %env_base, i64 0)
  br label %done

plain:
  %fnptr2 = load ptr, ptr %fv
  %fp2 = bitcast ptr %fnptr2 to ptr (i64)*
  call void %fp2(i64 0)
  br label %done
```
Closure body (`main$1`) takes an explicit ctx parameter:
```llvm
define void @main$1(ptr %env_base, i64 %v) {
entry:
  %i_ptr = load ptr, ptr %env_base
  store i64 %v, ptr %i_ptr
  ret void
}
```

## `__llgo_closure_const$...`
These are constant closure objects in read‑only data. They:
1) carry env without heap allocation, and
2) provide stable, deduplicated closure identities for type metadata (map hash/eq helpers, etc.).

## Discussion: Alternative Layout (difference only)
Alternative split layout:
```go
type closure struct {
  fn ptr
  data *struct { hasCtx bool; env ... }
}
```
Differences vs `*funcval` (no value judgment):
- The closure value is always two words, and the `data` object contains `hasCtx` + `env`.
- Call sites need one extra indirection (`data`) to reach `env`.
- Constant closures can be represented as two constants (`closure` + `data`), instead of a single `funcval` constant.
- The `data` object may be heap allocated or embedded in other objects depending on escape/placement.

## Covered Scenarios
Plain funcs, captured closures, method values/expressions, interface method values, varargs, `go/defer`, C callbacks.

### DIFF
--- a/internal/ctxreg/ctxreg.go
+++ b/internal/ctxreg/ctxreg.go
@@ -31,6 +31,10 @@ func ReserveFlags(goarch string) []string {
 		return []string{"-msse2"}
 	default:
 		// Use target-feature to reserve the register across backends.
-		return []string{"-Xclang", "-target-feature", "-Xclang", "+reserve-" + info.Name}
+		// Suppress warning about clobbering reserved registers in inline asm.
+		return []string{
+			"-Xclang", "-target-feature", "-Xclang", "+reserve-" + info.Name,
+			"-Wno-inline-asm",
+		}
 	}
 }
--- a/internal/llgen/normalize.go
+++ b/internal/llgen/normalize.go
@@ -63,9 +63,9 @@ func init() {
 		writeAsm := fmt.Sprintf(tmpl.writeFmt, info.Name)
 		readAsm := fmt.Sprintf(tmpl.readFmt, info.Name)
 
-		// WriteCtxReg pattern: call void asm sideeffect "<write_asm>", "r,~{reg}[,~{memory}]"(ptr %...)
+		// WriteCtxReg pattern: call void asm sideeffect "<write_asm>", "r[,~{reg}][,~{memory}]"(ptr %...)
 		writePattern := fmt.Sprintf(
-			`call void asm sideeffect "%s", "r,~\{%s\}(?:,~\{memory\})?"\(ptr [^)]+\)`,
+			`call void asm sideeffect "%s", "r(?:,~\{%s\})?(?:,~\{memory\})?"\(ptr [^)]+\)`,
 			writeAsm, info.Name,
 		)
 		writeCtxRegPatterns = append(writeCtxRegPatterns, regexp.MustCompile(writePattern))
--- a/ssa/expr.go
+++ b/ssa/expr.go
@@ -371,6 +371,18 @@ func (b Builder) WriteCtxReg(val Expr) {
 	if ctxRegCallerSaved(b.Prog.target.GOARCH) {
 		parts = append(parts, "~{memory}")
 	}
+	if len(parts) > 1 {
+		seen := make(map[string]bool, len(parts))
+		uniq := parts[:0]
+		for _, p := range parts {
+			if seen[p] {
+				continue
+			}
+			seen[p] = true
+			uniq = append(uniq, p)
+		}
+		parts = uniq
+	}
 	constraints := strings.Join(parts, ",")
 	asm := llvm.InlineAsm(ftype, writeAsm, constraints, true, false, dialect, false)
 	b.impl.CreateCall(ftype, asm, []llvm.Value{casted.impl}, "")
--- a/cl/_testcall/closureparam/out.ll
+++ b/cl/_testcall/closureparam/out.ll
@@ -67,7 +67,7 @@ _llgo_2:                                          ; preds = %_llgo_1
   %14 = load i64, ptr %13, align 4
   %15 = load ptr, ptr %1, align 8
   %16 = getelementptr i8, ptr %1, i64 16
-  call void asm sideeffect "mov x26, $0", "r,~{x26},~{memory}"(ptr %16)
+  call void asm sideeffect "mov x26, $0", "r,~{x26}"(ptr %16)
   %17 = call i64 %15(i64 %14)
   %18 = extractvalue %"github.com/goplus/llgo/runtime/internal/runtime.Slice" %3, 0
   %19 = extractvalue %"github.com/goplus/llgo/runtime/internal/runtime.Slice" %3, 1
@@ -85,7 +85,7 @@ _llgo_3:                                          ; preds = %_llgo_1
 
 define i64 @"github.com/goplus/llgo/cl/_testcall/closureparam.main$2"(i64 %0) {
 _llgo_0:
-  %1 = call ptr asm sideeffect "mov $0, x26", "=r,~{memory}"()
+  %1 = call ptr asm sideeffect "mov $0, x26", "=r"()
   %2 = load { ptr }, ptr %1, align 8
   %3 = extractvalue { ptr } %2, 0
   %4 = load i64, ptr %3, align 4
--- a/cl/_testcall/freeclosure/out.ll
+++ b/cl/_testcall/freeclosure/out.ll
@@ -21,7 +21,7 @@ _llgo_0:
   %0 = call ptr @"github.com/goplus/llgo/cl/_testcall/freeclosure.main$1"(i64 3)
   %1 = load ptr, ptr %0, align 8
   %2 = getelementptr i8, ptr %0, i64 16
-  call void asm sideeffect "mov x26, $0", "r,~{x26},~{memory}"(ptr %2)
+  call void asm sideeffect "mov x26, $0", "r,~{x26}"(ptr %2)
   %3 = call i64 %1(i64 2)
   ret void
 }
@@ -45,7 +45,7 @@ _llgo_0:
 
 define i64 @"github.com/goplus/llgo/cl/_testcall/freeclosure.main$1$1"(i64 %0) {
 _llgo_0:
-  %1 = call ptr asm sideeffect "mov $0, x26", "=r,~{memory}"()
+  %1 = call ptr asm sideeffect "mov $0, x26", "=r"()
   %2 = load { ptr }, ptr %1, align 8
   %3 = extractvalue { ptr } %2, 0
   %4 = load i64, ptr %3, align 4
--- a/cl/_testcall/go/out.ll
+++ b/cl/_testcall/go/out.ll
@@ -232,7 +232,7 @@ _llgo_2:                                          ; preds = %_llgo_0
 
 define void @"github.com/goplus/llgo/cl/_testcall/go.main$1"() {
 _llgo_0:
-  %0 = call ptr asm sideeffect "mov $0, x26", "=r,~{memory}"()
+  %0 = call ptr asm sideeffect "mov $0, x26", "=r"()
   %1 = load { ptr }, ptr %0, align 8
   %2 = extractvalue { ptr } %1, 0
   %3 = load i64, ptr %2, align 4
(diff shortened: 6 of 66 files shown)


visualfc (approved):

internal/ctxreg/ctxreg.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// ReserveFlags returns clang flags to reserve the ctx register for a GOARCH.
+// We prefer LLVM backend reservation to avoid frontend limitations across targets.
+func ReserveFlags(goarch string) []string {
+	info := Get(goarch)
+	if info.Name == "" {
+		return nil
+	}
+	return []string{"-mllvm", "--reserve-regs-for-regalloc=" + info.Name}
```
zhouguangyuan0718: This option is for aarch64 only, it will not work for other archs. reference to [here](https://github.com/llvm/llvm-project/blob/58f851dfb66dcd0af89d0e2da483a358c3643114/llvm/lib/Target/AArch64/AArch64Subtarget.cpp#L63).  The option to reserve register in llvm is not same for all archs. For aarch64 and riscv64,  should use command option "-mattr=+reserve-x26"  ,reference to  [here](https://github.com/llvm/llvm-project/blob/9865171e24961d9ae85d7183d5f52c44b82a9c58/llvm/test/CodeGen/AArch64/arm64-platform-reg.ll#L21) and [here](https://github.com/llvm/llvm-project/blob/aa1d2693c25622ea4a8ee2b622ba2a617e18ef88/llvm/test/CodeGen/RISCV/reserved-regs.ll#L52). Or set the "target-feature" attr to the function, like:
```
define void @reserve_x26() "target-features"="+neon,+reserve-x26"
```

Unfortunately, as I know,  similar feature is not supported for x86......

internal/llgen/normalize.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Asm instruction templates per architecture family (register name is %s placeholder)
+	// These define the LLVM IR asm string format for context register operations
+	asmTemplates = map[string]struct {
+		writeFmt string // format for write asm, e.g. "mov \\$0, %%%s" -> "mov \$0, %r12"
+		readFmt  string // format for read asm, e.g. "mov %%%s, \\$0" -> "mov %r12, \$0"
+	}{
+		"amd64":   {writeFmt: "mov \\$0, %%%s", readFmt: "mov %%%s, \\$0"},
+		"386":     {writeFmt: "mov \\$0, %%%s", readFmt: "mov %%%s, \\$0"},
+		"arm64":   {writeFmt: "mov %s, \\$0", readFmt: "mov \\$0, %s"},
```
zhouguangyuan0718, line 41: Reference to the previous [comment](https://github.com/goplus/llgo/pull/1568/changes#r2746377016). If ignore the support issue for amd64 and use "+reserve-x26" to reserve the register. Then use an llvm intrinsic to access the register is better.  Reference to [`llvm.read_register` and `llvm.write_register`](https://llvm.org/docs/LangRef.html#llvm-read-register-llvm-read-volatile-register-and-llvm-write-register-intrinsics). It will avoid the redundant "move instruction" and the sideeffect.

ssa/expr.go
```diff
@@ -298,6 +300,118 @@ func (b Builder) InlineAsmFull(instruction, constraints string, retType Type, ex
...
+		read:    "mov %%%s, $0",
+		dialect: llvm.InlineAsmDialectATT,
+	},
+	"386": {
+		write:   "mov $0, %%%s",
+		read:    "mov %%%s, $0",
+		dialect: llvm.InlineAsmDialectATT,
+	},
+	"arm64": {
+		write:   "mov %s, $0",
```
zhouguangyuan0718, line 321: Same with https://github.com/goplus/llgo/pull/1568#discussion_r2746446660

internal/ctxreg/ctxreg.go
cpunion: Thanks for the note — I rechecked this across targets with both clang and llc.

- -mllvm --reserve-regs-for-regalloc=<reg> works across all tested platforms. Ubuntu amd64 CI already passes with this.
- -mattr=+reserve-<reg> is llc-only. clang does not accept -mattr (it always reports “unknown argument”).
- -ffixed-* and +reserve-* are not portable: some targets reject them outright, others ignore the feature.

So the most portable path is still -mllvm --reserve-regs-for-regalloc=.... -mattr only makes sense when driving llc
directly, and -ffixed/+reserve are target‑specific.

Evidence (x86_64/i386):

# x86_64: -ffixed (clang)
clang -target x86_64-unknown-linux-gnu -x ir -c x86_readwrite.ll -o /tmp/a.o -ffixed-r12

clang: error: unknown argument '-ffixed-r12'; did you mean '-ffixed-r19'?

# x86_64: +reserve (clang)
clang -target x86_64-unknown-linux-gnu -x ir -c x86_readwrite.ll -o /tmp/a.o \
  -Xclang -target-feature -Xclang +reserve-r12

'+reserve-r12' is not a recognized feature for this target (ignoring feature)
fatal error: error in backend: Invalid register name global variable

# i386: -ffixed (clang)
clang -target i386-unknown-linux-gnu -x ir -c i386_readwrite.ll -o /tmp/a.o -ffixed-esi

clang: error: unknown argument: '-ffixed-esi'

# i386: +reserve (clang)
clang -target i386-unknown-linux-gnu -x ir -c i386_readwrite.ll -o /tmp/a.o \
  -Xclang -target-feature -Xclang +reserve-esi

'+reserve-esi' is not a recognized feature for this target (ignoring feature)

# x86_64: --reserve-regs-for-regalloc (clang, simple IR)
clang -target x86_64-unknown-linux-gnu -x ir -c simple.ll -o /tmp/a.o \
  -mllvm --reserve-regs-for-regalloc=r12

(no error)

Extra note on -mattr:
-mattr=+reserve-* is accepted by llc (e.g. AArch64/RISC‑V), but clang does not accept -mattr at all:

clang: error: unknown argument: '-mattr=+reserve-x26'

internal/llgen/normalize.go
cpunion: 1) llvm.read_register / llvm.write_register support

- x86_64: not supported (fails with Invalid register name global variable)
- i386: works (tested with esi)
- AArch64: works (tested with x26)
- RISC‑V 32/64: works (tested with x27)
- Other targets (ARMv7/MIPS/PPC/S390x/WASM/AVR/Xtensa, etc.) are untested; LangRef warns allocatable GPR support is limited.

2) IR becomes longer + needs memory barrier

- Inline asm has ~{memory} baked in, which is a compiler‑level barrier.
- If we switch to intrinsics, we must add explicit memory clobber to keep the same ordering.
- Minimal equivalent is 2 IR instructions per read/write:

```llvm
  call void asm sideeffect "", "~{memory}"()
  %v = call i64 @llvm.read_volatile_register.i64(metadata !0)

  call void asm sideeffect "", "~{memory}"()
  call void @llvm.write_register.i64(metadata !0, i64 %val)
  (So read+write becomes 4 IR instructions instead of 2.)
```

Because x86_64 does not support the intrinsics, and the intrinsic path adds extra IR (requires memory barriers), it is safer
and simpler to keep the current inline asm approach rather than switching.

zhouguangyuan0718: 1. Support on x86 should be a separate issue, now only [these register](https://github.com/llvm/llvm-project/blob/91dad7e7cab490693bbfeaf7a4c454bd779ae82f/llvm/lib/Target/X86/X86ISelLowering.cpp#L28664-L28671) are supported.  For AArch64/Riscv, if using target-feature to reserve the register, `llvm.read_register / llvm.write_register` should work. 
2. Why we need the memory barrier? IMO, the final assembly instructions order may be scheduled different with the IR order, but should not broke the semantics. Or are there some other reason? 
And the instruction count in IR is not same with the assembly instructions. If using inline asm, the asm code in the inlineASM expression will be kept in the final asm. But if using intrinsic, it will not generate extra instruction in asm code, just using the register directly. 
Which mean:
#### InlineASM:
```
  %1 = call ptr asm sideeffect "mov $0, x26", "=r,~{memory}"()
  %2 = load { ptr }, ptr %1, align 8
```
==> 
```
mov x0, x26 // x0 is selected by llvm, maybe other, but this mov instruction will not be eliminated
ldr x1, [x0]
```

#### intrinsic:
```
  %1 = call i64 @llvm.read_register.i64(metadata !7)
  %2 = load { ptr }, ptr %1, align 8

  !7 = !{!"x26"}
```
==> 
```
ldr x1, [x26]   // x1 is selected by llvm, maybe other, but there is no mov instruction, the reserved register can be accessed directly.
```

cpunion: Updated. For caller-saved x86, inline asm uses a memory clobber; callee-saved targets do not.

Issue #1559, opened 2026-01-14
unexpect panic: promoted embedded method missing in itab for interface assertions


luoliwoshang:
`cl/_testgo/interface`

```go
package main

import (
	"github.com/goplus/llgo/cl/_testdata/foo"
)

type Game1 struct {
	*foo.Game
}

type Game2 struct {
}

func (p *Game2) initGame() {
}

func main() {
	var g1 any = &Game1{&foo.Game{}}
	var g2 any = &Game2{}
	v1, ok := g1.(foo.Gamer)
	println("OK", v1, ok)
	if ok {
		v1.Load()
	}
	v2, ok := g2.(foo.Gamer)
	println("FAIL", v2, ok)
}
```
expect
```
cl/_testgo/interface on  expect-output-tests [$!?] via 🐹 v1.24.10 
❯ go run in.go
OK (0x12b94ffc8,0x14000104000) true
load
FAIL (0x0,0x1400005c727) false
```

got
```bash
cl/_testgo/interface on  expect-output-tests [$!?] via 🐹 v1.24.10 
❯ llgo run in.go
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
WARNING: Using LLGO root for devel: /Users/zhangzhiyang/Documents/Code/goplus/llgo
OK (0x102c83f90,0x102c81fc0) true
panic: runtime error: invalid memory address or nil pointer dereference

[0x02A4EAF8 github.com/goplus/llgo/runtime/internal/runtime.init#4$1+0x38, SP = 0x48]
[0x8AB5C624 _sigtramp+0x9, SP = 0x38]
[0x02A50A8C command-line-arguments.main+0x1b, SP = 0x70]
exit status 2
```

PR #1576, opened 2026-01-19
fix(runtime): make NewItab tolerate unreachable IFn entries

luoliwoshang:
Fixed https://github.com/goplus/llgo/issues/1559

 ## Summary
  - Align itab construction with Go’s behavior for “matched but unreachable/stripped” methods: `findMethod` now returns `(matched, fn)`. Only the “not matched” case clears the itab; the “matched but fn is nil” case writes a placeholder so the itab stays intact.
  - Rationale: Go emits full IFn/Tfn for all methods (including promoted/unexported) at compile time, then the linker deadcode pass rewrites unreachable entries to -1 (which maps to `unreachableMethod`). Thus interface assertions succeed, and calls to unreachable methods panic at
  call time, not at assertion time. LLGo previously zeroed IFn/Tfn early for cross-package unexported methods; `findMethod` would match name/signature but return nil, causing `itab.fun[0]` to be cleared and yielding a nil deref after a “successful” assertion. This change keeps the
  itab populated when the method matches, even if the function pointer was stripped for size/DCE reasons, deferring failure to the call site.
  - Reset `ssa/abitype.go` and LL test outputs (`cl/_testgo/interface/out.ll`, `cl/_testgo/abimethod/out.ll`) to main to avoid unrelated diffs.

### DIFF
--- a/runtime/internal/runtime/z_face.go
+++ b/runtime/internal/runtime/z_face.go
@@ -133,7 +133,7 @@ func NewItab(inter *InterfaceType, typ *Type) *Itab {
 		data := (*uintptr)(c.Advance(ptr, int(itabHdrSize)))
 		mthds := u.Methods()
 		for i, m := range inter.Methods {
-			matched, fn := findMethod(mthds, m)
+			fn, matched := findMethod(mthds, m)
 			if !matched {
 				ret.fun[0] = 0
 				break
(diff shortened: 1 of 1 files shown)


runtime/internal/runtime/z_face.go
```diff
@@ -147,18 +153,22 @@ func NewItab(inter *InterfaceType, typ *Type) *Itab {
 	return ret
 }
 
-func findMethod(mthds []abi.Method, im abi.Imethod) abi.Text {
+// findMethod searches for an interface method in the type's method set.
+// Returns (matched, fn) where:
+//   - matched: true if a method with matching name and type signature exists
+//   - fn: the function pointer (may be nil even when matched=true for promoted unexported methods)
+func findMethod(mthds []abi.Method, im abi.Imethod) (bool, abi.Text) {
```
xushiwei: swap `(bool, abi.Text)` to `(abi.Text, bool)`

luoliwoshang: @xgopilot adopt this suggestion!

PR #1596, opened 2026-02-02
refactor: remove init$patchDeps mechanism (no longer needed after PR #1450)

luoliwoshang:
## Summary

Since PR #1450 changed ABI types to compile-time constant data, the `init$after` lazy loading mechanism no longer exists. This means the initialization order issue that PR #1414 fixed is no longer possible.

### Before PR #1450
- Types were initialized at runtime via `init$after` with lazy loading (check null → create)
- Cross-package initialization order could cause types to be skipped
- Required `init$patchDeps` to ensure correct initialization order

### After PR #1450
- Types are compile-time constants embedded directly in LLVM IR
- No runtime initialization, no lazy loading, no order issues
- `init$patchDeps` mechanism is no longer needed

### Before (with `init$patchDeps`)

```llvm
define void @"go/build.init"() {
_llgo_1:
  store i1 true, ptr @"go/build.init$guard", align 1
  call void @"go/build.init$patchDeps"()    ; Separate function for dependency inits
  call void @"go/build.init$hasPatch"()     ; Only global var initialization
  ...
}

define void @"go/build.init$patchDeps"() {
  call void @"bufio.init"()
  call void @"bytes.init"()
  ; ... more dependency inits
  ret void
}

define void @"go/build.init$hasPatch"() {
  ; Only global variable initialization (guard check with inverted branches)
}
```

### After (simplified)

```llvm
define void @"go/build.init"() {
_llgo_1:
  store i1 true, ptr @"go/build.init$guard", align 1
  call void @"go/build.init$hasPatch"()     ; Contains BOTH dependency inits AND global vars
  ...
}

; No more init$patchDeps function!

define void @"go/build.init$hasPatch"() {
  ; Dependency init calls stay here (not extracted):
  call void @"bufio.init"()
  call void @"bytes.init"()
  ; ... more dependency inits
  ; Global variable initialization
}
```

### Diff

```diff
 define void @"go/build.init"() {
   store i1 true, ptr @"go/build.init$guard", align 1
-  call void @"go/build.init$patchDeps"()
   call void @"go/build.init$hasPatch"()
 }

-define void @"go/build.init$patchDeps"() {
-  call void @"bufio.init"()
-  call void @"bytes.init"()
-  ; ...
-  ret void
-}

 define void @"go/build.init$hasPatch"() {
+  call void @"bufio.init"()      ; dependency inits now stay here
+  call void @"bytes.init"()
+  ; ...
   ; global variable initialization
 }
```

## Code Changes

This PR removes:
- `init$patchDeps` function generation
- `initDepsFnNameOfHasPatch()` helper
- `findFirstNonInitInstruction()` helper
- `createInitDepsFunction()` method
- Related `.init$patchDeps` handling in `notInit()`

## Related PRs
- #1450 - ssa: abi type global constant data
- #1414 - Fix: Correct initialization order for overlay packages (this PR reverts it)

### DIFF
--- a/cl/compile.go
+++ b/cl/compile.go
@@ -484,12 +484,6 @@ func (p *context) compileBlock(b llssa.Builder, block *ssa.BasicBlock, n int, do
 		p.debugParams(b, block.Parent())
 	}
 
-	// Handle init$hasPatch: split dependency init calls into separate function
-	if doModInit && p.state == pkgHasPatch {
-		baseName := strings.TrimSuffix(p.fn.Name(), "$hasPatch")
-		instrs = p.createInitDepsFunction(pkg, baseName, instrs)
-	}
-
 	if doModInit {
 		if pyModInit = p.pyMod != ""; pyModInit {
 			last = len(instrs) - 1
@@ -508,15 +502,7 @@ func (p *context) compileBlock(b llssa.Builder, block *ssa.BasicBlock, n int, do
 	isCgoCmacro := isCgoCmacro(fnName)
 	for i, instr := range instrs {
 		if i == 1 && doModInit && p.state == pkgInPatch { // in patch package but no pkgFNoOldInit
-			name := p.fn.Name()
-			initFnNameOld := initFnNameOfHasPatch(name)
-			initPatchDepsFnName := initDepsFnNameOfHasPatch(name)
-
-			// Call deps function first
-			fnDeps := pkg.NewFunc(initPatchDepsFnName, llssa.NoArgsNoRet, llssa.InC)
-			b.Call(fnDeps.Expr)
-
-			// Then call hasPatch (global vars)
+			initFnNameOld := initFnNameOfHasPatch(p.fn.Name())
 			fnOld := pkg.NewFunc(initFnNameOld, llssa.NoArgsNoRet, llssa.InC)
 			b.Call(fnOld.Expr)
 		}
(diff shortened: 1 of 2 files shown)


luoliwoshang:
Closing this PR. After further analysis, the `init$patchDeps` mechanism is still required.

While PR #1450 removed the `init$after` type initialization (types are now compile-time constants), the `init$patchDeps` function serves a different critical purpose:

1. **Dependency package initialization**: It calls the `init()` functions of all dependency packages
2. **Global variable initialization order**: Ensures that dependency packages' global variables are initialized before the current package uses them

The `init$hasPatch` function uses **inverted guard logic** (branches are swapped at compile.go:309-310), so it executes when guard=true. The flow is:

```
overlay.init() {
    guard = false, enter _llgo_1
    set guard = true
    call init$patchDeps()  ← initializes all dependencies
    call init$hasPatch()   ← guard=true, enters _llgo_1 (inverted), initializes globals
}
```

Without `init$patchDeps`, dependency packages won't be initialized, causing runtime failures when accessing their global variables.

Sorry for the confusion!

luoliwoshang:
Reopening this PR. After re-analysis:

The `createInitDepsFunction` **extracts** dependency init calls from `init$hasPatch` into a separate `init$patchDeps` function.

If we remove this extraction:
- Dependency init calls **stay inside** `init$hasPatch`
- When `init$hasPatch` executes (guard=true due to inverted branches), it will:
  1. First call all dependency init functions
  2. Then initialize global variables

The execution is equivalent but simpler:

**Before (with init$patchDeps):**
```
overlay.init() → init$patchDeps() → init$hasPatch()
                      ↓                    ↓
               call dep inits      init global vars
```

**After (simplified):**
```
overlay.init() → init$hasPatch()
                      ↓
          call dep inits + init global vars (all inside)
```

Same functionality, cleaner code!

luoliwoshang:
@xushiwei

PR #1639, opened 2026-02-13
runtime: reflect.Copy

luoliwoshang:
## Summary
- add a minimal `reflect.Copy(dst, src Value) int` implementation in `runtime/internal/lib/reflect/value.go`
- add `typesMustMatch` helper used by `reflect.Copy`
- support Go-compatible `reflect.Copy` paths:
  - `dst`: `Array` / `Slice`
  - `src`: `Array` / `Slice`
  - special case: `src` is `String` when `dst` element kind is `Uint8`
- keep overlap-safe copy semantics via `memmove`
- add demo coverage in `_demo/go/reflectcopy/main.go` with both `expect` and `panic` checks

## Notes
- this PR intentionally keeps scope minimal for `reflect.Copy`
- `reflect.AppendSlice` remains unchanged (`todo`) in this PR

## Verification
Executed in `_demo/go/reflectcopy`:

```bash
go run main.go
llgo run main.go
```

Both produced identical output:

```text
expect slice-slice n ok
expect slice-slice dst ok
expect string-byte n ok
expect string-byte dst ok
expect array-slice n ok
expect array-slice dst ok
expect slice-array n ok
expect slice-array dst ok
panic dst-kind ok
panic array-unsettable ok
panic elem-mismatch ok
DONE
```

### DIFF
--- a/_demo/go/reflectcopy/main.go
+++ b/_demo/go/reflectcopy/main.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+func expect(name string, got, want any) {
+	if !reflect.DeepEqual(got, want) {
+		panic(fmt.Sprintf("expect %s: got %v, want %v", name, got, want))
+	}
+	fmt.Println("expect", name, "ok")
+}
+
+func expectPanic(name, contains string, fn func()) {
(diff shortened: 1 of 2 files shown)


luoliwoshang:
@xushiwei

PR #1642, opened 2026-02-15
ssa: fix loop defer draining

cpunion:
Fix llgo's defer lowering for defers executed inside loops.

- Ensure loop defers are recorded even for non-closure + no-args forms.
- Drain only loop-defer nodes (avoid popping unrelated defers), preventing panics like `time: Stop called on uninitialized Timer`.
- Classify defers in cyclic CFG blocks as loop defers (SCC-based) so nested/complex loops execute all defers.

Tests:
- Add regressions in `test/defer_test.go`.
- Strengthen `ssa/eh_loop_test.go`.
- Update `cl/_testdefer/gobuild/out.txt` to match new classification.

## Rationale / Notes

- `blocks`-only is not sufficient: on `goplus/main`, the loop-defer lowering in `ssa/eh.go` fails to record some loop defers (notably non-closure + no-args forms), so the drain loop has nothing to pop/run. In an isolated cherry-pick of only the SCC classification + tests, `go test ./ssa -run TestDeferInLoopIR` fails and the IR contains no `FreeDeferNode` for loop defers.
- This PR does not change `runtime/internal/runtime.Defer` (its layout stays the same). The only "structure" change is the compiler-emitted internal linked-list node layout stored in `Defer.Args` (adding a small header field `id uintptr`). Runtime never decodes these nodes; it only frees them via `FreeDeferNode`.
- The per-node `id` tag (or an equivalent discriminator) is required to preserve correct semantics when multiple defer statements inside the same loop share a single LIFO chain and their nodes are interleaved. Without a tag, the drainer cannot safely determine which static defer statement should decode/call a given node, and it risks decoding the wrong layout or draining unrelated defers.
- When a deferred call panics while draining loop defers, rethrow must resume at the drainer entry so the remaining nodes in the same frame are drained before unwinding to the next defer statement.

## Possible follow-up

- The IR generation for loop-drainers could be refactored to generate one drainer per contiguous run of `DeferInLoop` statements during `endDefer` (instead of accumulating state on `aDefer`), but the semantic requirements above (recording nodes, per-node dispatch, and rethrow-to-drain) remain.

### DIFF
--- a/ssa/eh.go
+++ b/ssa/eh.go
@@ -325,6 +325,14 @@ func (b Builder) Defer(kind DoAction, fn Expr, args ...Expr) {
 				return
 			}
 
+			// If a deferred call panics while draining loop defers, we must keep
+			// draining the remaining nodes in this frame before unwinding to the
+			// next defer statement. To achieve this, rethrow should resume at the
+			// current stmt block (which is a valid indirectbr destination),
+			// re-entering the drain loop with the next node already popped.
+			drainEntry := b.blk
+			drainEntryAddr := drainEntry.Addr()
+
 			prog := b.Prog
 			condBlk := b.Func.MakeBlock()
 			exitBlk := b.Func.MakeBlock()
@@ -365,6 +373,7 @@ func (b Builder) Defer(kind DoAction, fn Expr, args ...Expr) {
 				b.If(match, caseBlks[i], nextBlk)
 
 				b.SetBlockEx(caseBlks[i], AtEnd, true)
+				b.Store(self.rethPtr, drainEntryAddr)
 				b.callDefer(self, c.typ, c.fn, c.args)
 				b.Jump(condBlk)
 			}


cpunion:
Addressed review notes:

- ssa/eh.go: add loopDrainerGenerated and generate the loop-defer drainer once per contiguous run of DeferInLoop statements (reset when hitting DeferAlways/DeferInCond), plus guard against empty loopCases.
- ssa/eh.go: update the pseudo-code node layout comment to include the new id field.
- ssa/eh.go: replace the offset==2 proxy with an explicit fn.kind==vkClosure check.
- ssa/eh.go: add doc comment for loopDeferCase.
- cl/blocks/block.go: clarify loop (traversal marker) vs inLoop (SCC semantics).

Re dispatch perf: ssa.Builder doesn't currently expose a switch helper (it's commented out), so keeping the linear chain for now.

cpunion:
Follow-up fix: loop-defer drainer must survive rethrow.

`loopDrainerGenerated` is fine for codegen dedup, but a deferred call can panic while draining (e.g. recover+re-panic). In that case we must keep draining remaining loop nodes in the same frame.

Fix: in the DeferInLoop drainer, store `rethPtr` to the current stmt block address before each `callDefer`, so a rethrow resumes at the drainer stmt (valid indirectbr dest) and continues draining with the next node already popped.

This makes `TestPanicCrossTwoFunctionsRecover` pass under `llgo test`.

PR #1655, opened 2026-02-26
runtime: fix darwin syscall errno detection on amd64

### LINKED ISSUE #1654: llgo on macOS amd64: concurrent temp-file operations panic with `bad file descriptor`
## Background
In PR https://github.com/goplus/llcppg/pull/662, the CI matrix fails only on `macos amd64` with `llgo 0.12.2`.
Other matrix targets (`linux arm64/amd64`, `macos arm64`) do not show this failure.

Failing test:
```txt
--- FAIL: TestEnd2End/sqlite3/3.49.1 (19.45s)
```

## Failure Symptom
From the failed artifact (`macos-15-intel-log.zip`):

```txt
panic: close /var/folders/.../compose_1723566520.h: bad file descriptor

[0x02352E6D github.com/goplus/llcppg/cmd/llcppg.main+0x28, SP = 0x70d]
[0x02355A67 main+0x4, SP = 0x27]
[0x0F9AC530 start+0x5, SP = 0xbf0]
```

This aligns with the `TestEnd2End/sqlite3/3.49.1` failure window.

## Environment
- OS: `macos-15-intel` (`darwin/amd64`)
- llgo: `v0.12.2-0.20260210235731-9c8b6b3df1ac` (`darwin/amd64`)
- Go (control): `go1.23.6 darwin/amd64`

## Minimal Reproduction
This reproducer does **not** depend on llcppg logic. It only does concurrent `CreateTemp + Close + Remove`.

```go
package main

import (
	"fmt"
	"os"
	"sync"
)

const (
	goroutines = 2
	iterations = 1000
)

func worker(n int, errs chan<- error) {
	for i := 0; i < n; i++ {
		f, err := os.CreateTemp("", "compose_*.h")
		if err != nil {
			errs <- err
			return
		}
		name := f.Name()
		if err := f.Close(); err != nil {
			errs <- fmt.Errorf("close %s: %w", name, err)
			return
		}
		if err := os.Remove(name); err != nil {
			errs <- err
			return
		}
	}
}

func main() {
	errs := make(chan error, goroutines)
	var wg sync.WaitGroup

	for i := 0; i < goroutines; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			worker(iterations, errs)
		}()
	}

	wg.Wait()
	close(errs)

	for err := range errs {
		if err != nil {
			panic(err)
		}
	}

	fmt.Printf("ok goroutines=%d iterations=%d\n", goroutines, iterations)
}
```

Run:
```bash
# Fails under llgo (concurrent case)
llgo run repro.go
# panic: close ... bad file descriptor

# Passes under Go toolchain
go run repro.go
# ok goroutines=2 iterations=1000
```

## Additional Repro (single-thread, deterministic)
The issue is reproducible even without concurrency:

```go
package main

import (
	"fmt"
	"os"
	"path/filepath"
	"syscall"
)

func main() {
	p := filepath.Join(os.TempDir(), fmt.Sprintf("open-fail-%d.tmp", os.Getpid()))
	_ = os.WriteFile(p, []byte("x"), 0600)
	defer os.Remove(p)

	fd, err := syscall.Open(p, syscall.O_CREAT|syscall.O_EXCL|syscall.O_RDWR, 0600)
	fmt.Printf("fd=%d hex=%#x err=%v\n", fd, uint64(fd), err)
}
```

Observed output:

```bash
# go run
fd=4294967295 hex=0xffffffff err=file exists

# llgo run
fd=4294967295 hex=0xffffffff err=<nil>
```

This means the syscall failed (invalid fd `0xffffffff`) but `err` was lost under llgo.

## Root Cause Analysis (likely)
- `os.CreateTemp` internally relies on `OpenFile(..., O_CREATE|O_EXCL, ...)`.
- Under contention, `open` may return `EEXIST` (normal behavior).
- With llgo on `darwin/amd64`, this failure can be interpreted as `err=nil` while returning invalid fd (`0xffffffff`).
- The invalid fd then propagates into `*os.File`, and later operations (`Write` / `Stat` / `Close`) fail with `bad file descriptor`.
- Concurrency amplifies filename-collision opportunities, so the bug appears mostly in concurrent test runs.

Likely implementation mismatch in llgo syscall lowering:
- errno is derived only when `r1 == ^uintptr(0)`.
- on this path, failing C calls may produce a 32-bit `-1` shape (`0x00000000ffffffff`) rather than full-width `^uintptr(0)` (`0xffffffffffffffff`) on amd64.
- then errno is not captured, causing `err=nil` on failure.

## Expected Behavior
No panic; temp file descriptors should remain valid and close cleanly under concurrency.

## Actual Behavior
On `llgo v0.12.2` + `darwin/amd64`, concurrent temp-file usage can panic with `bad file descriptor`.

## Impact on llcppg
End-to-end concurrent tests (notably `sqlite3`) can fail on this specific toolchain/platform combination.

luoliwoshang:
Fixes #1654 #1661

## Summary
This PR addresses two Darwin/amd64-specific syscall trampoline issues in LLGo and adds regressions.

1. `open`/`openat`/`fcntl`/`ioctl` path:
- Fixes syscall trampoline ABI width/sign handling so failed libc calls are preserved as full-width `-1` (`^uintptr(0)`) and errno can be detected reliably.

2. `fdopendir`/`readdir_r` (and `closedir`) path:
- Adds Darwin wrappers to avoid darwin/amd64 `INODE64` symbol-name mismatches in trampoline resolution.

Also includes validation-oriented Darwin/amd64 CI/demo updates already in this branch.

### A) Width/sign mismatch in trampoline return path
On darwin/amd64, failed `open`-family libc calls could be surfaced as `0x00000000ffffffff` (narrowed/sign-lost) instead of full-width `^uintptr(0)`.  
That can bypass LLGo's `r1 == -1` errno detection path.

Observed effect:
- `syscall.Open(..., O_CREAT|O_EXCL, ...)` on an existing path may return invalid fd with `err=nil`.
- `os.CreateTemp` then misses `EEXIST` and can fail unexpectedly instead of retrying.

### B) Darwin/amd64 INODE64 symbol mismatch
For some directory APIs, Darwin/amd64 may resolve symbols via `...$INODE64` naming conventions.  
Direct trampoline name mapping is not always compatible, so wrapper indirection is needed.

### Runtime wrappers / trampoline mapping
- `runtime/internal/clite/os/_os/os.c`
  - Uses machine-word style return handling for Darwin trampoline wrappers.
  - Preserves failure sentinel via `(uintptr_t)(intptr_t)ret`.
- `runtime/internal/clite/os/_os/os_darwin.c`
  - Adds:
    - `llgo_fdopendir`
    - `llgo_closedir`
    - `llgo_readdir_r`
- `runtime/internal/clite/os/os_darwin.go`
  - Includes `_os/os_darwin.c` in `LLGoFiles`.
- `cl/instr.go`
  - Extends Darwin trampoline remap table:
    - `fdopendir -> llgo_fdopendir`
    - `closedir -> llgo_closedir`
    - `readdir_r -> llgo_readdir_r`
  - Adds rationale comments clarifying:
    - varargs/ABI width issue (`open/openat/fcntl/ioctl`)
    - INODE64-name compatibility (`fdopendir/readdir_r`, plus `closedir` for call-chain consistency)

### Regression demos
- `_demo/go/sysopen-1654/main.go`
  - Deterministic `syscall.Open` repro for `O_CREAT|O_EXCL` existing-file behavior.
- `_demo/go/createtemp-1654/main.go`
  - Concurrent stress repro to ensure `os.CreateTemp` no longer hits intermittent bad-fd behavior from missed errno.

### CI / demo updates in this branch
- `.github/actions/setup-go/action.yml`
- `.github/workflows/llgo.yml`
- `_demo/c/asmfullcall/asmfullcall_darwin_amd64.go`
- `_demo/c/asmfullcall/asmfullcall_darwin_arm64.go` (rename from generic darwin file)

## Validation
Manually validated on Darwin/amd64 with LLGo demos:

- `cd _demo/go/sysopen-1654 && llgo run .`
  - Expected and observed: existing-file `O_CREAT|O_EXCL` returns `fd == -1`, `err == EEXIST`.

- `cd _demo/go/createtemp-1654 && llgo run .`
  - Repeated runs no longer show intermittent `bad file descriptor` caused by missed errno.

## Notes
- No intended functional behavior change for non-Darwin targets.
- `closedir` wrapper is included to keep the directory syscall call-chain consistent with the same wrapper/ABI strategy.

### DIFF
--- a/.github/workflows/llgo.yml
+++ b/.github/workflows/llgo.yml
@@ -36,7 +36,10 @@ jobs:
   llgo:
     needs: download-model
     continue-on-error: true
-    timeout-minutes: 30
+    # macOS Intel is much slower than macOS ARM in the "Test demos" step.
+    # Observed: macOS ARM about 2m49s vs macOS Intel about 10m26s.
+    # Keep a larger budget to avoid timeout-related cancellations.
+    timeout-minutes: 60
     strategy:
       matrix:
         os:


MeteorsLiu (approved):

.github/workflows/llgo.yml
```diff
@@ -41,6 +41,7 @@ jobs:
       matrix:
         os:
           - macos-latest
+          - macos-15-intel
```
luoliwoshang, line 47: current setup-go will download unexpect go version for `macos-15-intel`,make current `setup-go` action ensure the `macos-15-intel` with correct go os-arch version,also need other step work normally such as darwin/arm64 linux/amd64
```bash
Prepare all required actions
Run ./.github/actions/setup-go
Downloading Go from: https://go.dev/dl/go1.24.2.darwin-arm64.tar.gz
```
@xgopilot

luoliwoshang, line 47: @xgopilot make compatibility for `./_demo/c/asmfullcall` for darwin amd64

cpunion (approved):

cpunion (approved):

PR #1666, opened 2026-02-28
ssa: preserve export functions with llvm.compiler.used

zhouguangyuan0718:
## Summary

Use `llvm.compiler.used` to preserve exported symbols during LTO.

## Changes

- preserve exported symbols with `@llvm.compiler.used`
- stop relying on reading or merging pre-existing `@llvm.used` entries
- mark preserved symbols explicitly from export handling
- keep the preservation logic local to the current package state
- update SSA coverage for the `llvm.compiler.used` behavior

## Notes

This version treats `llvm.compiler.used` as its own preservation list and does not depend on any existing `llvm.used` contents in the module.

### DIFF
--- a/ssa/decl.go
+++ b/ssa/decl.go
@@ -211,6 +211,9 @@ func (p Package) NewFuncEx(name string, sig *types.Signature, bg Background, has
 	if instantiated {
 		fn.SetLinkage(llvm.LinkOnceAnyLinkage)
 	}
+	if p.isPreservedName(name) {
+		p.markLLVMUsed(fn)
+	}
 	ret := newFunction(fn, t, p, p.Prog, hasFreeVars)
 	p.fns[name] = ret
 	return ret
--- a/ssa/package.go
+++ b/ssa/package.go
@@ -436,7 +436,9 @@ func (p Program) NewPackage(name, pkgPath string) Package {
 		mod: mod, Prog: p, vars: gbls, fns: fns,
 		pyobjs: pyobjs, pymods: pymods, strs: strs,
 		di: nil, cu: nil, glbDbgVars: glbDbgVars,
-		export: make(map[string]string),
+		export:         make(map[string]string),
+		preserveSyms:   make(map[string]struct{}),
+		llvmUsedValues: make([]llvm.Value, 0, 4),
 	}
 	ret.abi.Init(pkgPath, uintptr(p.ptrSize), (*goProgram)(unsafe.Pointer(p)))
 	return ret
@@ -701,7 +703,9 @@ type aPackage struct {
 	NeedPyInit  bool
 	NeedAbiInit bool // need load all abi types for reflect make type
 
-	export map[string]string // pkgPath.nameInPkg => exportname
+	export         map[string]string   // pkgPath.nameInPkg => exportname
+	preserveSyms   map[string]struct{} // set of exported symbol names
+	llvmUsedValues []llvm.Value
 }
 
 type Package = *aPackage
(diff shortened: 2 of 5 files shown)


ssa/decl.go
```diff
@@ -211,6 +211,9 @@ func (p Package) NewFuncEx(name string, sig *types.Signature, bg Background, has
 	if instantiated {
 		fn.SetLinkage(llvm.LinkOnceAnyLinkage)
 	}
+	if _, ok := p.exportNames[name]; ok {
+		fn.SetDLLStorageClass(llvm.DLLExportStorageClass)
+	}
```
luoliwoshang: I agree that //export symbols need extra marking to avoid being dropped by LTO, but I have a question: why choose to uniformly set dllexport?

According to LLVM LangRef, dllexport is a DLL storage class with Windows/COFF (and XCOFF) export-interface semantics, and it does not seem to be a general-purpose “keep-alive” marker. Why wasn’t ExternalLinkage chosen here to preserve the symbol instead?

https://llvm.org/docs/LangRef.html#dll-storage-classes

ssa/package.go
```diff
@@ -714,10 +719,33 @@ func (p Package) SetExport(name, export string) {
...
+	if !p.llvmUsed.IsNil() {
+		p.llvmUsed.EraseFromParentAsGlobal()
+	}
+	init := llvm.ConstArray(elemTyp, p.llvmUsedValues)
+	global := llvm.AddGlobal(p.mod, init.Type(), "llvm.compiler.used")
+	global.SetInitializer(init)
+	global.SetLinkage(llvm.AppendingLinkage)
+	global.SetSection("llvm.metadata")
+	p.llvmUsed = global
+}
```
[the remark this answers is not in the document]
luoliwoshang: @zhouguangyuan0718 this suggestion seems good , maybe have a look~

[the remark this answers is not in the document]
zhouguangyuan0718: done

ssa/package.go
```diff
@@ -701,7 +703,10 @@ type aPackage struct {
 	NeedPyInit  bool
 	NeedAbiInit bool // need load all abi types for reflect make type
 
-	export map[string]string // pkgPath.nameInPkg => exportname
+	export         map[string]string   // pkgPath.nameInPkg => exportname
+	preserveSyms   map[string]struct{} // set of exported symbol names
```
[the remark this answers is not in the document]
zhouguangyuan0718: Only support function export

ssa/package.go
```diff
@@ -701,7 +703,10 @@ type aPackage struct {
...
+	llvmUsed       llvm.Value
```
luoliwoshang: After the https://github.com/goplus/llgo/pull/1666/files#r2875486295 comment ~ we remove the package level `llvm.compiler.used` 's value erase and reset logic,maybe current we not need this field ~

luoliwoshang (commented):
This PR correctly handles preservation for the `//export` path, but there is another export entry path that should be preserved as well: the `cPkg` auto-export path.

ssa/package.go
```diff
@@ -714,10 +719,37 @@ func (p Package) SetExport(name, export string) {
...
+	p.llvmUsedValues = append(p.llvmUsedValues, llvm.ConstBitCast(v, elemTyp))
+}
+
+func (p Package) MaterializePreserveSyms() {
+	if len(p.llvmUsedValues) == 0 || !p.llvmUsed.IsNil() {
+		return
+	}
+	elemTyp := p.Prog.VoidPtr().ll
+	init := llvm.ConstArray(elemTyp, p.llvmUsedValues)
+	global := llvm.AddGlobal(p.mod, init.Type(), "llvm.compiler.used")
```
luoliwoshang, line 742: Small suggestion: would `llvm.used` be a better fit here?  
Ref (LLVM LangRef on `llvm.used` vs `llvm.compiler.used`):  
https://llvm.org/docs/LangRef.html#the-llvm-used-global-variable  
https://llvm.org/docs/LangRef.html#the-llvm-compiler-used-global-variable  

Using `llvm.compiler.used` to keep symbols alive during compiler/optimization passes is correct; just noting that with a later link stage, `llvm.used` may better match final-retention semantics.

luoliwoshang: https://llvm.org/docs/LangRef.html#the-llvm-compiler-used-global-variable
```
The @llvm.compiler.used directive is the same as the @llvm.used directive, except that it only prevents the compiler from touching the symbol. On targets that support it, this allows an intelligent linker to optimize references to the symbol without being impeded as it would be by @llvm.used.
```

zhouguangyuan0718: `llvm.used` will cause undefined symbol, seems like it has some other efffect during link time. llvm.compiler.used can work well, and when enable LTO, we only need it work in "compile time"(Which actually run in linker).

ssa/package.go
zhouguangyuan0718: Good point, done

ssa/package.go
```diff
@@ -701,7 +703,9 @@ type aPackage struct {
 	NeedPyInit  bool
 	NeedAbiInit bool // need load all abi types for reflect make type
 
-	export map[string]string // pkgPath.nameInPkg => exportname
+	export         map[string]string   // pkgPath.nameInPkg => exportname
+	preserveSyms   map[string]struct{} // set of exported symbol names
```
luoliwoshang, line 707: current we may only need keep the function ,so current `preserveSyms` and `export` seem reduant,maybe use `p.export[name]` is sufficient
```go
if p.isPreservedName(name) {
		p.markLLVMUsed(fn)
	}
```

zhouguangyuan0718: The key in export is the original name, the value is exportname, we need collect the export name.

luoliwoshang: Sorry, I mixed up the mapping. In export, the key is the original symbol and the value is the exported symbol; NewFuncEx gets the final emitted symbol name, so p.export[name] is not a direct lookup here. Using preserveSyms as a set of exported names makes sense. Thanks for clarifying.

ssa/package.go
luoliwoshang: Got it

PR #1671, opened 2026-03-02
fix(esp32c3): include INIT/PREINIT/FINI array sections in ESP image builder

### LINKED ISSUE #1672: esp32c3 emulator startup panic: INIT_ARRAY section dropped from ESP image (affects multiple cases)
## Summary

`esp32c3-basic` emulator startup panic is not limited to `fncall`.

Affected reproducible cases (at least):

- `./cl/_testdata/fncall`
- `./cl/_testdata/untyped`
- `./cl/_testdata/varinit`

All of them fail with `Instruction access fault` very early in startup (`PC=0x0`, `RA` inside `__libc_init_array`).

## Why this happens

These paths use `libc` startup, which executes constructor arrays (`__libc_init_array`).

The previous ESP `elf -> img` packer only included `SHT_PROGBITS` sections. But for these cases, constructor-related data is in `SHT_INIT_ARRAY` (or related array section types), so it gets dropped from the image.

Then startup walks missing constructor entries and traps at `PC=0x0`.

## Repro

```bash
llgo run -a -target=esp32c3-basic -emulator ./cl/_testdata/fncall
llgo run -a -target=esp32c3-basic -emulator ./cl/_testdata/untyped
llgo run -a -target=esp32c3-basic -emulator ./cl/_testdata/varinit
```

Typical panic pattern:

```text
Guru Meditation Error: Core 0 panic'ed (Instruction access fault)
PC      : 0x00000000
RA      : 0x403802xx  (__libc_init_array)
```

## Extra evidence

For `untyped`/`varinit`, old image had 4 segments and missed init-array payload; with fix, image has 5 segments and includes an extra segment at init-array address (e.g. `0x403803bc,len:0x4`).

## Not viable approach

Forcing output `.rodata` type to `SHT_PROGBITS` is not a safe fix when input contains `.init_array`; linker reports type mismatch:

```text
ld.lld: error: section type mismatch for .rodata
>>> ...(.init_array): SHT_INIT_ARRAY
>>> output section .rodata: SHT_PROGBITS
```

## Expected fix direction

Update ESP image packing to include constructor-array section types as loadable data, not only `SHT_PROGBITS`:

- `SHT_INIT_ARRAY`
- `SHT_PREINIT_ARRAY`
- `SHT_FINI_ARRAY`

Candidate fix PR: #1671.

luoliwoshang:
Fixes #1672

## Summary

This PR fixes the ESP32-C3 emulator startup fault for `./cl/_testdata/fncall` and also restores the linker script layout introduced in PR #1435.

Changes:

1. `esp -> img` packing now includes constructor-array related section types:
- `SHT_INIT_ARRAY`
- `SHT_PREINIT_ARRAY`
- `SHT_FINI_ARRAY`

2. `targets/esp32-riscv.app.elf.ld` is restored to the pre-`765ab4b9` structure (separate `.preinit_array` / `.init_array` / `.fini_array`, no longer merged into `.rodata`).

3. Keep `.rodata ALIGN(4)` to satisfy ESP image block alignment requirements after restoring separate sections.

## Why this is needed

We are using `libc` startup on this path (`__libc_init_array`), so constructor arrays must be present in the generated image.

The old workaround in PR #1435 (merging `init_array` into `.rodata`) worked in some cases only because `.rodata` also had other regular constants, and the final output section stayed effectively packable as regular data.

In `fncall`, there are effectively no extra constants around that constructor-table path, so section typing/layout no longer behaves like those earlier cases; relying on the merge hack is fragile.

Now that `esp -> img` handles `INIT/PREINIT/FINI_ARRAY` directly, we can safely restore the linker script to the cleaner separate-array layout.

## Validation

Using local build from current source:

```bash
PATH="/Users/heulucklu/project/llgo/.cache/qemu/bin:$PATH" \
  /tmp/llgo-local run -a -target=esp32c3-basic -emulator ./cl/_testdata/fncall
```

Result: emulator boots and exits normally (no `Instruction access fault`).

### DIFF
--- a/_demo/embed/test_esp32c3_startup.sh
+++ b/_demo/embed/test_esp32c3_startup.sh
@@ -6,8 +6,8 @@
 #
 # Verifies:
 # 1. _start uses newlib's __libc_init_array (not TinyGo's start.S)
-# 2. .init_array section is merged into .rodata section
-# 3. .rodata (including .init_array) is included in BIN file
+# 2. __init_array_start symbol is present in ELF
+# 3. __init_array_start is included in BIN load segments
 
 set -e
 
@@ -26,7 +26,7 @@ trap cleanup EXIT
 
 # Check if esptool.py is installed
 # esptool.py is required to parse ESP32-C3 BIN file format and verify
-# that .rodata segment (containing .init_array) is included in the firmware
+# that constructor-related data is included in the firmware
 if ! command -v esptool.py &> /dev/null; then
     echo "✗ FAIL: esptool.py not found"
     echo "Please install: pip3 install esptool==5.1.0"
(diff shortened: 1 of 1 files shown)


targets/esp32-riscv.app.elf.ld
```diff
@@ -22,34 +22,12 @@ SECTIONS
     PROVIDE (etext = .);
   } > iram_seg
 
-  .rodata    :
+  .rodata ALIGN(4) :
```
MeteorsLiu, line 25: Could you please provide the documentation source for why rodata must be ALIGN(4)? I checked the `newlib` for esp target, [https://github.com/espressif/newlib-esp32/blob/007ee8a2eb5b60a69dfbdb4bf2399413b46a0b6b/libgloss/riscv/esp/app.elf.ld#L26](https://github.com/espressif/newlib-esp32/blob/007ee8a2eb5b60a69dfbdb4bf2399413b46a0b6b/libgloss/riscv/esp/app.elf.ld#L26), they did not apply ALIGN(4) to the entire rodata section. 

My concern is that applying 4-byte alignment to the entire rodata section without verifying it against the documentation might lead to some unexpected behavior.

luoliwoshang: Without .rodata ALIGN(4), .rodata may start at a non-4-byte-aligned address in our current build path.

  In our reproducible case, this is exactly what happens:

```bash
  load:0x40380396,len:0x18
  Invalid image block, can't boot.
  ets_main.c 333
```
  So the issue is not about changing .rodata contents; it is about guaranteeing the start address of the .rodata output section is 4-byte aligned.

  Also, .rodata ALIGN(4) only affects the section start location. It does not override or change alignment rules of individual input items inside .rodata (those still follow their own alignment
  requirements).

cpunion (approved):

luoliwoshang:
@xushiwei

PR #1680, opened 2026-03-04
fix: normalize 32-bit slice bounds

### LINKED ISSUE #1679: esp32c3-basic emulator: cl/_testgo/reader hangs and does not auto-exit
### Problem
When running `cl/_testgo/reader` with ESP32-C3 emulator target, the process hangs and does not auto-exit.

### Reproduce
```bash
cd cl/_testgo/reader
llgo run -a -target=esp32c3-basic -emulator .
```

### Actual behavior
QEMU boots and reaches `entry 0x40380000`, then the run does not finish automatically (hang).

### Expected behavior
The case should print expected output (`hello world (0x0,0x0)`) and exit automatically.

### Scope
Observed during ESP32-C3 emulator regression checks for PR #1567.

luoliwoshang:
## Summary
- normalize `low/high/max` to target `int` width in SSA `Slice` before calling runtime helpers
- add ESP serial startup regression output for `int64` string slicing (`slice64 ok`)
- make smoke assertion compare expected output tail by N lines (instead of only the last line)

## Root Cause
On 32-bit targets, passing wider bounds (for example `int64`) directly into runtime slice helpers with `int` semantics can cause incorrect behavior (including hangs/no auto-exit).

## Changes
- `ssa/datastruct.go`
  - in `Builder.Slice`, apply `fitIntSize` to non-nil `low/high/max` before helper calls
- `_demo/embed/test-esp-serial-startup.sh`
  - add `int64` slicing regression output in generated `main`
  - update smoke matcher to validate expected tail lines

## Verification
- `go install ./cmd/llgo`
- `_demo/embed/test-esp-serial-startup.sh`
  - ESP32-C3: pass (tail includes `Hello World` + `slice64 ok`)
  - ESP32: pass for this smoke case with local QEMU installed

Fixes #1679

### DIFF
--- a/_demo/embed/test-esp-serial-startup.sh
+++ b/_demo/embed/test-esp-serial-startup.sh
@@ -1,40 +1,16 @@
 #!/bin/bash
-# ESP serial targets smoke test (build + emulator run only).
+# ESP serial targets smoke test (emulator run only).
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TEMP_DIR="$SCRIPT_DIR/.test_tmp_$$"
 
-ESP32C3_PREFIX="esp32c3_smoke"
-ESP32_PREFIX="esp32_smoke"
 CASE_ROOT="$SCRIPT_DIR/testdata/esp32-serial"
 
 cleanup() {
     rm -rf "$TEMP_DIR"
 }
 trap cleanup EXIT
-
-build_target() {
-    local target="$1"
-    local prefix="$2"
-    local label="$3"
-    local test_go="$4"
-
-    echo "==> Building for $label target ($target): ELF + BIN..."
-    llgo build -target="$target" -o "$prefix" -obin "$test_go"
-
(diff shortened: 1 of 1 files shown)


_demo/embed/test-esp-serial-startup.sh
```diff
@@ -84,6 +100,14 @@ import "github.com/goplus/lib/c"
 
 func main() {
 	c.Printf(c.Str("Hello World\n"))
+	s := "hello"
+	var idx int64 = 1
```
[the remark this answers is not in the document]
luoliwoshang: already change to compare all line bbe037c4d29055fc6814b5a35c87a807410cf70f

_demo/embed/test-esp-serial-startup.sh
```diff
@@ -93,10 +117,10 @@ echo ""
 echo "=== ESP Serial Smoke Tests: Build + Emulator Run ==="
```
cpunion: Split test cases and test runner

luoliwoshang: done

PR #1706, opened 2026-03-12
build: reject GOOS/GOARCH cross requests without -target

### LINKED ISSUE #553: Run failed when set GOOS and GOARCH even cross-compile is ignored
When I try to make support for `llgo` to build wasm file. It occurs errors but not only in wasm case:

1. In my Mac M1, clone repo,  run `go build -v github.com/goplus/llgo/cmd/llgo && ./llgo build -o main-binary main.go`, everything works fine.
2. If I try `GOOS=linux GOARCH=arm64 ./llgo build -o main-binary main.go` (use other os and arch pair is same), It shows errors:

```
c/c_linux.go:25:11: undefined: FilePtr
c/c_linux.go:28:12: undefined: FilePtr
c/c_linux.go:31:12: undefined: FilePtr
...
```
all llgo/c packages show same type error with undefined type.

3. I found the errors are generated from  https://github.com/goplus/llgo/blob/main/internal/packages/load.go#L390

I think we can make it runnable, ignore GOOS and GOARCH to build local target binary; or provide error message to show cross-compile is not supported.

cpunion:
## Summary
- reject unsupported host cross-compilation requests that only set `GOOS`/`GOARCH`
- fail early in `internal/build` with a clear diagnostic instead of falling through to linker or asm errors
- add unit coverage for host-matching and unsupported cross requests

## Rationale
LLGo does not currently support host cross-compilation just by setting `GOOS`/`GOARCH`.
Before this change, those requests fell through into deeper build and link stages and failed with misleading downstream errors.
This change makes that unsupported path explicit.

Fixes #553

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -495,7 +495,7 @@ func validateHostCrossRequest(goos, goarch, target string) error {
 	if goos == runtime.GOOS && goarch == runtime.GOARCH {
 		return nil
 	}
-	return fmt.Errorf("cross-compilation via GOOS/GOARCH requires -target (host=%s/%s requested=%s/%s)", runtime.GOOS, runtime.GOARCH, goos, goarch)
+	return fmt.Errorf("GOOS/GOARCH cross-compilation is not supported (host=%s/%s requested=%s/%s)", runtime.GOOS, runtime.GOARCH, goos, goarch)
 }
 
 func needLink(pkg *packages.Package, mode Mode) bool {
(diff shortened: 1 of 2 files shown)


cpunion:
Closing this PR.

The change made unsupported `GOOS`/`GOARCH` host cross requests fail earlier with a clearer error, but that is not the right product direction for #553.

`GOOS=linux GOARCH=arm64` should eventually be supported. The current problem is that this path still falls through into downstream toolchain/linker failures on `main`; it is not a case we want to reject permanently.

Issue #553 has been updated with the current reproduction/status instead.

PR #1710, opened 2026-03-12
runtime: implement basic FuncForPC support

cpunion:


cpunion:
Closing this for now. The current FuncForPC fix is incomplete, and the more robust direction is to first remove/avoid closure stubs so function values expose real code addresses. We can revisit FuncForPC after that work lands.

PR #1728, opened 2026-03-18
ssa: emit go semantic metadata

luoliwoshang:
## Background

Ordinary `call/ref` reachability only sees direct symbol references, but in Go the question of which methods must be kept also depends on additional semantics such as:

- interface conversions
- interface method calls
- `MethodByName`
- conservative reflection handling

An ordinary reference graph alone cannot answer:

- which concrete types enter the interface dispatch domain
- which interface method slots are actually demanded
- which method names are requested by name
- which sites must fall back to conservative reflection handling

This PR is not the end of the full pipeline. It is one implementation stage of the proposal: it emits these Go semantic facts as `llgo.xxx` metadata during `ssa/cl`, and adds stable readback for them.

This PR corresponds to the producer / readback portion of:

- `llgo#1727: Proposal: Semantic Pruning of Unreachable Methods from a Global Graph`

This PR **only covers** the producer / readback layer of the proposal. It **does not include**:

- whole-program analysis after metadata aggregation
- global method liveness driven by `OrdinaryEdges` / `TypeChildren`

### 1. Emit Go semantic metadata during `ssa/cl`

This PR adds and wires up the following named metadata:

- `!llgo.useiface`
- `!llgo.useifacemethod`
- `!llgo.interfaceinfo`
- `!llgo.methodinfo`
- `!llgo.usenamedmethod`
- `!llgo.reflectmethod`

They represent:

- `useiface`
  - which concrete types enter the interface semantic domain once an owner becomes reachable
- `useifacemethod`
  - which interface method demands are produced once an owner becomes reachable
- `interfaceinfo`
  - the complete method set of an interface
- `methodinfo`
  - the method-table slots of a concrete type, plus `MType / IFn / TFn`
- `usenamedmethod`
  - which method names are requested exactly through `MethodByName`
- `reflectmethod`
  - which owners must fall back to conservative reflection handling

### 2. Define the metadata encoding baseline

The current encoding is row-oriented to keep emission and later aggregation simple:

- `llgo.interfaceinfo`
  - one row per interface method
- `llgo.methodinfo`
  - one row per method slot

In particular:

- `methodinfo` is emitted only when a type actually has method slots
- unexported method names are normalized with package qualification
- `MethodInfo` keeps `Index / MethodSig / IFn / TFn`, matching the needs of later analysis

### 3. Distinguish the emission ownership of `MethodInfo` and `InterfaceInfo`

Both are inputs to later analysis, but they intentionally use different emission strategies in the current implementation.

#### `MethodInfo`

`MethodInfo` is tied to the method-table layout of a **concrete type**. Its ownership is stronger, and duplicate emission is much noisier.

So this PR tightens its emission:

- for ordinary imported, non-generic named types
  - `llgo.methodinfo` is no longer emitted redundantly at use sites
- their method-slot metadata is emitted only by the defining package

Generic instances remain conservatively allowed at use sites for now:

- current LLGo still materializes generic instance methods primarily at use sites
- a concrete imported generic named instance may only be instantiated and compiled in the current package
- if imported generic instances were forced into definition-site-only emission now, metadata could be missing even though the instance methods were materialized locally

So the current policy is:

- ordinary imported, non-generic named types
  - deduplicated; emitted only by the defining package
- imported generic instances
  - still conservatively allowed at use sites, prioritizing completeness

#### `InterfaceInfo`

`InterfaceInfo` records the **complete method set of an interface**. In principle it also looks like definition-site information, but under current LLGo interface lowering, named interfaces are often erased early into their underlying `*types.Interface`. That makes it unreliable to recover a stable named owner from the definition-side type materialization path.

So the current implementation chooses:

- emit `InterfaceInfo` at the **use site**
- specifically, along the path that already produces the interface method demand

This is intentional because:

- that path always has the relevant interface shape in hand
- duplicate `InterfaceInfo` contributions from multiple modules remain semantically safe after whole-program merge
- this is more reliable than forcing a definition-site-only rule under the current lowering model

In other words, this PR intentionally treats the two differently:

- `MethodInfo`
  - heavier and more strongly bound to concrete type ownership, so duplicate emission is reduced aggressively
- `InterfaceInfo`
  - definition-site ownership is unstable under current lowering, so use-site emission is preferred to keep the input complete

### 4. Add `internal/semmeta`

This PR adds `internal/semmeta`, which is responsible for:

- reading `llgo.xxx` metadata back from a single `llvm.Module`
- folding it into semantic `ModuleInfo`
- owning the metadata protocol on both the write side and the read side

This layer only handles semantic metadata. It does not handle:

- ordinary LLVM reference graphs
- `TypeChildren`
- whole-program aggregation
- DCE analyze / rewrite

### 5. Use the new metadata read API from `goplus/llvm`

This PR also switches to the metadata read API that has already landed in `goplus/llvm`, removing the temporary cgo-based reader previously used by tests.

The current code directly uses:

- `Module.NamedMetadataOperands`
- `Value.MDNodeOperands`
- `Value.MDString`

to read named metadata back from `llvm.Module`.

### 6. Add `llgen.GenModuleFrom`

This PR adds `GenModuleFrom` to `internal/llgen`, so callers can directly obtain an `llvm.Module`.

That makes it possible to use the same `llvm.Module` for both:

1. obtaining IR text
2. reading semantic metadata back

without a stringify-then-reparse detour.

### 7. Build a stable semantic view for metadata

The readback result is not a raw row dump. It is a stable semantic view grouped by meaning, for example:

- `UseIface`
- `UseIfaceMethod`
- `InterfaceInfo`
- `MethodInfo`
- `UseNamedMethod`
- `ReflectMethod`

This captures semantic content rather than LLVM metadata node numbering or row layout details.

## Current Scope

This PR only covers the producer / readback layer of the proposal. It does not include:

- whole-program DCE analysis after metadata aggregation
- global method liveness driven by `OrdinaryEdges` / `TypeChildren`

In other words, this PR establishes the input contract and observability foundation required by the later algorithm.

## Representative Coverage

This PR covers several important semantic categories:

- `ifaceconv`
  - interface conversions driving `useiface`
- `reader` / `invoke` / `interface`
  - interface method demands
- `embedunexport` / `geometry1370`
  - interface methods with unexported names
- `reflectmk`
  - `Method`, `MethodByName`, and conservative reflection
- `abimethod`
  - named / anonymous / promoted methods and embedded method sets
- `cursor` / `reflectconv` / `abinamed`
  - large method-slot sets and more complex type coverage

## Follow-up Work

Later PRs will continue from here with:

1. whole-program aggregation of `OrdinaryEdges`, `TypeChildren`, and semantic metadata
2. the analysis that computes `type symbol -> live method indexes`


ssa/mdtest/metadata.go
luoliwoshang, line 1: will use goplus/llvm instead

luoliwoshang: already use goplus/llvm [ssa: use llvm metadata read APIs](https://github.com/goplus/llgo/pull/1728/commits/1fe4623a067c4ac40633489a0f762c6abdda5bdf)

luoliwoshang:
@xgopilot review

cl/_testgo/abimethod/meta-expect.txt
```diff
@@ -0,0 +1,201 @@
...
+
+[MethodInfo]
+*_llgo_bytes.Buffer:
+    0 Available _llgo_func$ETeB8WwW04JEq0ztcm-XPTJtuYvtpkjIsAc0-2NT9zA bytes.(*Buffer).Available bytes.(*Buffer).Available
+    1 AvailableBuffer _llgo_func$Z_-7GWzB37LCYRTQLsSYmEihg_hqBK8o_GbT88pqnPY bytes.(*Buffer).AvailableBuffer bytes.(*Buffer).AvailableBuffer
+    2 Bytes _llgo_func$Z_-7GWzB37LCYRTQLsSYmEihg_hqBK8o_GbT88pqnPY bytes.(*Buffer).Bytes bytes.(*Buffer).Bytes
+    3 Cap _llgo_func$ETeB8WwW04JEq0ztcm-XPTJtuYvtpkjIsAc0-2NT9zA bytes.(*Buffer).Cap bytes.(*Buffer).Cap
+    4 Grow _llgo_func$VZ-8VPNF1RaLICwxc1Ghn7BbgyFX3v762OCdx127EkA bytes.(*Buffer).Grow bytes.(*Buffer).Grow
+    5 Len _llgo_func$ETeB8WwW04JEq0ztcm-XPTJtuYvtpkjIsAc0-2NT9zA bytes.(*Buffer).Len bytes.(*Buffer).Len
+    6 Next _llgo_func$d4kMA_oCkLwnd1j8nVlv1hwRarEVuCIrDCpnHhDz9UY bytes.(*Buffer).Next bytes.(*Buffer).Next
```
luoliwoshang: This MethodInfo entry is expected: these are the promoted methods of the anonymous `type struct{ m int; *bytes.Buffer }`, inherited from the embedded
  `*bytes.Buffer`. The potentially redundant part is the separate `*_llgo_bytes.Buffer` MethodInfo row in this module, since the bytes package should
  already emit that type’s method table in its own module.

luoliwoshang: Updated in 2e945f4b.

  llgo.methodinfo emission is now gated so we no longer emit duplicate rows for imported non-generic named types at use sites. That keeps the
  promoted-method entries for the local anonymous type, but drops the redundant *_llgo_bytes.Buffer row from this module. Generic instances remain
  conservatively allowed for now, since current LLGo may still materialize their methods at the use site.

cl/_testgo/cgocfiles/meta-expect.txt
```diff
@@ -0,0 +1,4 @@
+[UseIface]
+github.com/goplus/llgo/cl/_testgo/cgocfiles.main:
+    _llgo_github.com/goplus/llgo/cl/_testgo/cgocfiles._Ctype_int
+    _llgo_string
```
luoliwoshang: This UseIface entry is expected. Although _Ctype_int does not have methods and will not contribute any MethodInfo-driven method reachability, it is still boxed into any here via fmt.Println(r), so it correctly enters the UseIface set:

```go
  r := C.test_structs(...)
  fmt.Println(r)
```

cl/_testgo/cgofull/meta-expect.txt
```diff
@@ -0,0 +1,18 @@
...
+    *_llgo_github.com/goplus/llgo/cl/_testgo/cgofull._Ctype_struct___sFILE
+    *_llgo_github.com/goplus/llgo/cl/_testgo/cgofull._Ctype_struct__object
+github.com/goplus/llgo/cl/_testgo/cgofull.runPy$2:
+    *_llgo_github.com/goplus/llgo/cl/_testgo/cgofull._Ctype_struct___sFILE
+    *_llgo_github.com/goplus/llgo/cl/_testgo/cgofull._Ctype_struct__object
+github.com/goplus/llgo/cl/_testgo/cgofull.runPy$3:
+    *_llgo_github.com/goplus/llgo/cl/_testgo/cgofull._Ctype_struct___sFILE
+    *_llgo_github.com/goplus/llgo/cl/_testgo/cgofull._Ctype_struct__object
+github.com/goplus/llgo/cl/_testgo/cgofull.triggerC2func:
+    _llgo_string
```
luoliwoshang: This mismatch is caused by platform-specific cgo type names rather than a semantic metadata regression. On macOS, FILE* is materialized as *_Ctype_struct___sFILE, while on glibc/Linux it becomes *_Ctype_struct__IO_FILE. Both correspond to the same
  C.stderr/C.stdout uses in runPy, so this meta-expect.txt should be treated as platform-dependent and skipped for now rather than compared literally across platforms.

see  https://github.com/goplus/llgo/actions/runs/24411453642/job/71309102954?pr=1728

luoliwoshang: skip at 572da2ddbb992484266aa67c52b0a46d2ebc47fe

cl/_testgo/reflect/meta-expect.txt
```diff
@@ -0,0 +1,94 @@
...
+
+[InterfaceInfo]
+_llgo_reflect.Type:
+    Align _llgo_func$ETeB8WwW04JEq0ztcm-XPTJtuYvtpkjIsAc0-2NT9zA
+    AssignableTo _llgo_func$Kxk9fspGkjXcoNWf2ucHG1vOQ5VHxVtYionfm-DnvWE
+    Bits _llgo_func$ETeB8WwW04JEq0ztcm-XPTJtuYvtpkjIsAc0-2NT9zA
+    CanSeq _llgo_func$YHeRw3AOvQtzv982-ZO3Yn8vh3Fx89RM3VvI8E4iKVk
+    CanSeq2 _llgo_func$YHeRw3AOvQtzv982-ZO3Yn8vh3Fx89RM3VvI8E4iKVk
+    ChanDir _llgo_func$JO3khPIbANSMBmoN6P7ybYAeUBd3Gv6toVUqNeE7qbE
+    Comparable _llgo_func$YHeRw3AOvQtzv982-ZO3Yn8vh3Fx89RM3VvI8E4iKVk
```
luoliwoshang, line 17: I think InterfaceInfoshould follow the same ownership rule asMethodInfo: for ordinary imported non-generic named
  interfaces, the full interface method set should be emitted by the defining package, not duplicated at each use site. In
  this case, reflect.Typeshould only contributeUseIfaceMethodrows here; the fullInterfaceInfo(_llgo_reflect.Type)belongs in
  thereflect package module.

luoliwoshang: fixed at 3b8bbed2086528859be22a24865abc5b79f23551

zhouguangyuan0718:
这部分我不太建议新增meta-expect.txt这种类型的用例，和之前的out.ll问题一样，针对性不足，并且会受其他部分变化的影响。我建议可以新增一些有针对性的测试用例，复用同样的LITTEST机制来做检查，而不是普遍性的测试大量代码的输出。

luoliwoshang:
> 这部分我不太建议新增meta-expect.txt这种类型的用例，和之前的out.ll问题一样，针对性不足，并且会受其他部分变化的影响。

明白，对于每个cl/testxxx 都输出meta-expect.txt 会导致噪音过多，这里预期会修改为针对性的用例验证metadata的产出。但如果emitter输出逻辑、abi.Type名称稳定其实这个回归也不应该变化，如果abi.Type名称变化这里更新其实是预期的～

> 我建议可以新增一些有针对性的测试用例，复用同样的LITTEST机制来做检查，而不是普遍性的测试大量代码的输出。

这里复用LITTEST机制 指的是预期在cl/testxxx下面新增对应的in.go，然后标记llvm module 的 metadata的节点产出进行匹配么；
因为这部分想回归的其实是metadata的产出稳定，而不是llvm.module 那里的metadata的节点布局，（后续如果涉及到缓存，可能会落到其他文件）所以这里还是会更倾向于对比的是序列化后的semantic module info ～ ，会更可读。

所以可能会更倾向于在cl/_testmeta下面增加 ifaceuse,interface,reflect等等的针对性用例，每个包内都是一个in.go，以及现在形状的meta-expect.txt.

zhouguangyuan0718:
> 明白，对于每个cl/testxxx 都输出meta-expect.txt 会导致噪音过多，这里预期会修改为针对性的用例验证metadata的产出。但如果emitter输出逻辑、abi.Type名称稳定其实这个回归也不应该变化，如果abi.Type名称变化这里更新其实是预期的～

主要是需要额外维护这些东西，比如因为其他原因，改了无关用例里的类型名称，或者顺序，字段之类的，还需要再去维护metadata的输出，这就和现在的偶尔改动需要刷新大量的out.ll的情况类似

> 所以可能会更倾向于在cl/_testmeta下面增加 ifaceuse,interface,reflect等等的针对性用例，每个包内都是一个in.go，以及现在形状的meta-expect.txt.

嗯嗯，我赞同这样的改法，主要是觉得把测试用例的关注点独立出来会比普遍检查输出会更好。

ssa/abitype.go
```diff
@@ -434,10 +435,73 @@ func (b Builder) abiUncommonMethods(t types.Type, mset *types.MethodSet) llvm.Va
...
+		return p.shouldEmitOwnedTypeMetadata(tt.Elem())
+	case *types.Named:
+		if ta := tt.TypeArgs(); ta != nil && ta.Len() != 0 {
+			return true
+		}
+		obj := tt.Obj()
+		if obj == nil || obj.Pkg() == nil {
+			return true
+		}
+		return abi.PathOf(obj.Pkg()) == p.Path()
```
zhouguangyuan0718: 这里有个疑问，看起来类型只有在定义的包中有使用的时候才会判定为true，那对于import的情况，如果在定义的包中没有使用，在其他包中才有使用，是不是会缺少这个type的MethodInfo？

ssa/interface.go
```diff
@@ -81,6 +82,35 @@ func (b Builder) Imethod(intf Expr, method *types.Func) Expr {
...
+			MType: semmeta.Symbol(mtypName),
+		},
+	})
+	// Emit named-interface shape at use sites as well. In current LLGo
+	// lowering, interface operations often erase a named interface to its
+	// underlying *types.Interface before abiType materialization, so relying
+	// only on definition/type-processing paths would miss rows such as
+	// "_llgo_foo/bar.IFmt" even though later whole-program analysis still
+	// needs that complete interface method set.
+	if _, ok := types.Unalias(intf.raw.Type).(*types.Named); ok {
```
zhouguangyuan0718: 匿名的interface是否也需要InterfaceInfo？

ssa/abitype.go
luoliwoshang: https://github.com/goplus/llgo/pull/1728#discussion_r3080607744 
确实是一个误优化！MethodInfo 当前是随 abiUncommonMethods/abiType 按需发射的，而不是定义包统一产出，所以不应该限制为只在 owner package emit。这里将会移除这个条件，和当前 InterfaceInfo 保持一致。

luoliwoshang: fixed at [d553c34](https://github.com/goplus/llgo/pull/1728/commits/d553c3408a8ed34ee8df0836a9bf76ef0eafb10a)[d553c34](https://github.com/goplus/llgo/pull/1728/commits/d553c3408a8ed34ee8df0836a9bf76ef0eafb10a)

ssa/interface.go
luoliwoshang: 已修改，匿名的interface也会需要Interface，否则后续全局视角计算下，类型是否实现接口在匿名接口的情况下会误判，已修改，并回归在 cl/_testmeta/interface_anonymous/in.go

luoliwoshang:
新的包级别的元数据Emit PR -> https://github.com/xgo-dev/llgo/pull/1874

PR #1734, opened 2026-03-20
cl: handle unnamed closure free vars in makeClosureCtx

visualfc:


cl/compile.go
```diff
@@ -265,7 +265,11 @@ func makeClosureCtx(pkg *types.Package, vars []*ssa.FreeVar) *types.Var {
 	n := len(vars)
 	flds := make([]*types.Var, n)
 	for i, v := range vars {
-		flds[i] = types.NewField(token.NoPos, pkg, v.Name(), v.Type(), false)
+		name := v.Name()
+		if name == "" {
+			name = "_"
+		}
```
luoliwoshang, line 271: see  `go@1.24/libexec/src/go/types/struct.go` ,when the `f.name != "_"` will not throw panic! check your comment ! @xgopilot 
```go
// NewStruct returns a new struct with the given fields and corresponding field tags.
// If a field with index i has a tag, tags[i] must be that tag, but len(tags) may be
// only as long as required to hold the tag with the largest index i. Consequently,
// if no field has a tag, tags may be nil.
func NewStruct(fields []*Var, tags []string) *Struct {
	var fset objset
	for _, f := range fields {
		if f.name != "_" && fset.insert(f) != nil {
			panic("multiple fields with the same name")
		}
	}
	if len(tags) > len(fields) {
		panic("more tags than fields")
	}
	s := &Struct{fields: fields, tags: tags}
	s.markComplete()
	return s
}
```

PR #1736, opened 2026-03-20
build:  Go Linktime-like Unreachable Method Pruning

luoliwoshang:
## Summary

Implements the **GlobalSummary consumer and method-liveness phase** described in [#1853](https://github.com/xgo-dev/llgo/issues/1853).

The package metadata producer, mmap-backed package summary cache, and global metadata view are provided by the preceding work in [#1989](https://github.com/xgo-dev/llgo/pull/1989). This PR consumes those facts and computes the concrete method slots that must remain live.

## What Changed

- Merge linked package metadata through `meta.GlobalSummary`.
- Seed analysis from executable entry roots:
  - `<entry package>.init`
  - `<entry package>.main`
  - runtime init when required.
- Propagate ordinary symbol reachability.
- Consume reachable `FuncDemand` facts:
  - interface conversions through `UseIface`
  - interface method calls through `UseIfaceMethod`
  - named method demands
  - conservative reflection demands.
- Propagate interface-use semantics through `TypeChildren`.
- Use `Ifaces`, `IfaceMethods`, and `MethodSlots` to establish complete concrete-type/interface implementation relationships.
- Run the fixed-point method-liveness analysis and return:
  `concrete type symbol -> live ABI method slot indexes`.
- Rewrite ABI type method tables in the entry module with strong overrides.
- Add analyzer, build integration, and LLVM rewrite golden coverage.

## Core Algorithm

For an ordinary interface-method call, a concrete method slot is kept only when these three conditions meet:

1. **Concrete type is in the interface-use domain**

   The concrete type is reached by a reachable `UseIface` fact. The domain is then propagated through `TypeChildren`, including the ABI pointer relationship needed to retain the corresponding pointer-receiver method set.

2. **The interface method demand is reachable**

   A reachable function produces a `UseIfaceMethod` demand such as `ReaderWriter.Read`. The analyzer records the demanded interface symbol, method name, and ABI method type.

3. **The concrete type completely implements the interface**

   The type has a complete implementation of that interface, and the candidate method slot matches the demanded interface method by both ABI method name and method type. The analyzer precomputes these implementation-to-slot relationships and runs the propagation to a fixed point.

In other words, interface-call liveness is the intersection of:

```text
types used as interface values
∩ reachable interface-method demands
∩ complete interface implementation/signature matches
```

Named-method demands and conservative reflection are separate conservative rules: they can keep matching method slots without requiring a specific interface-call intersection.

## ABI Rewrite

The analyzer does not mutate the original package modules. For every ABI type global with a method table, the rewrite pass copies the type initializer into the entry module and emits a same-name strong global. This strong definition overrides the original weak ABI type symbol at link time.

- Live slots copy the original method name, method type, `ifn`, and `tfn`.
- Dead slots copy the method name and method type, but replace `ifn` and `tfn` with the `runtime.unreachableMethod` declaration.
- Non-method ABI fields, method-table layout, and relevant global attributes are preserved.

The dead slot remains a valid non-zero method entry for runtime interface matching, while the original method implementation is no longer strongly referenced by the replacement method table.

## Scope

This PR implements the method-liveness consumer and the current ABI method-table rewrite. It does not add new metadata facts or change the package summary wire format.

The current integration is limited to executable builds. Plugin, shared-library, dynlink, arbitrary symbol rewriting, LLVM initializer overrides beyond ABI method tables, and further linker integration remain follow-up work.

## Enablement

Go method dropping is intentionally development-only and is disabled by default. Build a development `llgo` and pass:

```text
-deadcodedrop
```

The flag is unavailable in non-development builds. The existing Go global DCE path remains separate and takes precedence for full LTO unless explicitly disabled.

## Testing

- `go test ./internal/deadcode -count=1`
- `go test ./internal/dcepass -count=1`
- `go test ./internal/build -run 'TestApplyDeadcodeDropOverrides|TestDCEEntryRootCandidates' -count=1`

Related design and phase boundaries are documented in [#1853](https://github.com/xgo-dev/llgo/issues/1853).


zhouguangyuan0718:
使用//export导出的C函数看起来没有作为根 可能会漏掉  这个用例就能复现
```
package main

/*
extern int Callback(void);
static int callCallback(void) {
	return Callback();
}
*/
import "C"

type I interface {
	M() int
}

type T struct{}

func (T) M() int {
	return 7
}

//export Callback
func Callback() C.int {
	var v I = T{}
	return C.int(v.M())
}

func main() {
	println(C.callCallback())
}
```

internal/deadcode/analyze.go
```diff
@@ -0,0 +1,278 @@
...
+	for _, key := range d.methodImplKeys[id] {
+		if _, ok := d.ifaceMethod[key]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+func (d *pass) markMethod(method methodRef) {
+	d.liveSlots[method.owner] = append(d.liveSlots[method.owner], method.slot)
```
zhouguangyuan0718, line 245: 这里的method.slotInfo.MType应该不止是可达 应该还要usedInIface，这个例子可以构造问题
```
type T struct{}
type U struct{}

func (T) Make() U { return U{} }
func (U) Keep() int { return 42 }

func main() {
    u := reflect.ValueOf(T{}).MethodByName("Make").Call(nil)[0]
    println(u.MethodByName("Keep").Call(nil)[0].Int())
}
```

internal/dcepass/dcepass.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	methodsVal := init.Operand(init.OperandsCount() - 1)
+	if methodsVal.Type().TypeKind() != llvm.ArrayTypeKind {
+		return llvm.Value{}, llvm.Type{}, false
+	}
+	elemTy := methodsVal.Type().ElementType()
+	if elemTy.TypeKind() != llvm.StructTypeKind || elemTy.StructElementTypesCount() != 4 {
+		return llvm.Value{}, llvm.Type{}, false
+	}
+	if !strings.Contains(elemTy.StructName(), "runtime/abi.Method") {
```
zhouguangyuan0718: 这个感觉是不是应该用相等严格限定敏子 要不然可能会被外部代码误命中

internal/deadcode/analyze.go
luoliwoshang: 这里确实遗漏了该Case，已经修复，当某个Method被标记为可达时，其Mtype也被标记为UseIface，并补充该Case作为回归 c2786ec0f584340fa8187f0706b947039bcf92f9

internal/dcepass/dcepass.go
luoliwoshang: 已修改 [697033c](https://github.com/xgo-dev/llgo/pull/1736/commits/697033ca49076a990b6d5db1b02b71f92a98ccca)

luoliwoshang: 不过这里确实会和LTO碰到的问题类似，方法表不是独立的，所以会需要匹配格式和名称，后续我们或许可以考虑一下把方法表独立出去

luoliwoshang:
> 使用//export导出的C函数看起来没有作为根 可能会漏掉 这个用例就能复现

已增加 //export 函数为根的处理 [a1cb10b](https://github.com/xgo-dev/llgo/pull/1736/commits/a1cb10bfb2038d266e885566a7859c6578130cab)

zhouguangyuan0718 (approved):

PR #1739, opened 2026-03-23
ssa/abi: fix interfaceName check by private method pkg path

visualfc:
ssa/abi
- interfaceName check by private method pkg path
- structName check by private field pkg path
- rm Builder.Pkg

ssa
- change Package.abi to Program.abi


luoliwoshang (approved):
LGTM

cpunion (approved):

PR #1793, opened 2026-04-16
ssa: clamp float-to-int conversions

cpunion:
## Summary
- Clamp signed and unsigned float-to-int conversions before LLVM `fpto*` instructions.
- Avoid target-dependent overflow and NaN conversion results.
- Add Go/LLGO regression coverage for out-of-range and NaN conversions.

## Example
From `TestFloatToIntegerConversionBounds`:

```go
var f = math.Inf(1)
_ = int64(f)
_ = uint64(f)
_ = int64(math.NaN())
```

The test also covers values outside the target integer range.

## Without This PR
LLVM float-to-int overflow is poison/undefined for out-of-range values. LLGO could return unstable values or differ across platforms for NaN, infinities, and large finite floats.

## Verification
- `./dev/llgo.sh test ./test/go -run TestFloatToIntegerConversionBounds -count=1`
- `./dev/llgo.sh test ./test/go -count=1`
- `go test ./ssa -run TestFromTestrt/cast$ -count=1`
- `go test ./ssa -count=1`
- linux/amd64 container go1.24.11: `./dev/llgo.sh test ./test/go -run TestFloatToIntegerConversionBounds -count=1`
- linux/amd64 container go1.24.11: `go test ./ssa -run TestFromTestrt/cast$ -count=1`


cpunion:
Closing this PR because the current net diff is empty. The earlier float-to-int clamping changes were intentionally reverted to keep llgo aligned with gc: values that Go/gc does not define should not be made llgo-specific by this PR.

PR #1802, opened 2026-04-17
cl: skip unused array derefs in range

cpunion:
## Summary
- Skip emitting unused array dereference loads in range loops that only need array length.
- Preserve evaluation for expressions that have side effects.
- Add compiler and `./test/go` coverage for range over nil array pointers.

## Example
From `TestRangeOverNilArrayPointer`:

```go
var p *[3]int
sum := 0
for i := range p {
    sum += i
}
```

Range over an array pointer uses the array length and must not dereference `p` just to compute the loop count.

## Without This PR
LLGO can emit an unnecessary array load/deref for range loops that only need the array length. With nil array pointers this can panic or generate invalid code even though Go does not dereference the array for the loop count.

## Verification
- `go test ./test/go -run 'TestRangeOverNilArrayPointer' -count=1`
- `go run ./cmd/llgo test ./test/go -run 'TestRangeOverNilArrayPointer' -count=1`
- `go test ./cl -run 'TestCompile|TestRuntime' -count=1`

### DIFF
--- a/cl/compile.go
+++ b/cl/compile.go
@@ -644,56 +644,7 @@ func skipUnusedArrayDeref(v *ssa.UnOp) bool {
 	if _, ok := v.Type().Underlying().(*types.Array); !ok {
 		return false
 	}
-	return !valueHasArrayLenRangeEvalEffect(v.X, nil)
-}
-
-func valueHasArrayLenRangeEvalEffect(v ssa.Value, seen map[ssa.Value]bool) bool {
-	if seen[v] {
-		return false
-	}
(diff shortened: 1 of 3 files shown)


cpunion:
Updated in d75b249b8. Removed the recursive side-effect analysis and now only skips unused array deref instructions; SSA instruction order still preserves side effects such as calls. Also fixed the invalid nil-holder test case, added a call-evaluation regression case, and added cl unit coverage for skipUnusedArrayDeref.

PR #1806, opened 2026-04-17
ssa: align direct interface ABI

cpunion:
## Summary
- Align interface storage/extraction with Go direct-interface ABI rules.
- Box non-direct interface values and load non-direct interface data correctly.
- Add Go/LLGO regression coverage for reflect interface round-trips and scalar comparisons.

## Example
From `TestDirectInterfaceABIDecoding` and `TestReflectFieldInterfaceScalarComparison`:

```go
var x any = uintptr(123)
if reflect.ValueOf(x).Interface() != x { panic("bad interface round-trip") }
```

Scalar/direct-interface values and indirect interface values use different data representations and must be decoded accordingly.

## Without This PR
LLGO can mis-store or mis-load interface data. Reflect `Interface()` round-trips and scalar interface comparisons can read the wrong bits or compare the wrong representation, especially across amd64/arm64 ABI-sensitive paths.

## Verification
- `go test ./ssa -count=1`
- `go test ./test/go -count=1`
- `go run ./cmd/llgo test ./test/go -count=1`
- `./dev/docker.sh amd64 bash -lc 'go test ./test/go -run "TestDirectInterfaceABIDecoding|TestReflectFieldInterfaceScalarComparison" -count=1 && go run ./cmd/llgo test ./test/go -run "TestDirectInterfaceABIDecoding|TestReflectFieldInterfaceScalarComparison" -count=1'`
- `./dev/docker.sh arm64 bash -lc 'go test ./test/go -run "TestDirectInterfaceABIDecoding|TestReflectFieldInterfaceScalarComparison" -count=1 && go run ./cmd/llgo test ./test/go -run "TestDirectInterfaceABIDecoding|TestReflectFieldInterfaceScalarComparison" -count=1'`


cpunion:
Updated in b657a53c4. I checked the non-direct non-empty-interface path: `MakeInterface` still routes through `unsafeInterface`, which constructs the itab via `newItab`; the CI x509 failure was from `reflect.Value.Bool` reading indirect bool values using `flagAddr` instead of `flagIndir`. Added a native/llgo repro covering `reflect.DeepEqual(true, false)` and ASN.1 optional bool encoding, and revalidated the failing x509 tests in the linux/amd64 container.

PR #1812, opened 2026-04-17
test: add litgen tool for source IR checks

zhouguangyuan0718:
## Summary
- add a standalone `litgen` tool to generate and refresh source-embedded `// LITTEST` FileCheck directives from LLVM IR
- keep `gentests` focused on `out.ll`/`expect.txt` refreshes, but skip `out.ll` regeneration for directories whose source starts with `// LITTEST`
- document the two golden-refresh flows and add a minimal `_testrt/litdemo` example for `litgen`

## Testing
- `go test ./chore/gentests ./chore/litgen`
- `go test ./cl -run 'TestRunAndTestFromTestrt/litdemo$' -count=1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,7 +424,11 @@ cd llgo
 * [pysigfetch](https://github.com/goplus/hdq/tree/main/chore/pysigfetch): It generates symbol information by extracting information from Python's documentation site. This tool is not part of the `llgo` project, but we depend on it.
 * [llpyg](chore/llpyg): It is used to automatically convert Python libraries into Go packages that `llgo` can import. It depends on `pydump` and `pysigfetch` to accomplish the task.
 * [llgen](chore/llgen): It is used to compile Go packages into LLVM IR files (*.ll).
+* [gentests](chore/gentests): It refreshes the built-in golden test data under `cl/_test*`, including `out.ll` and `expect.txt`. Directories that use source-embedded `// LITTEST` checks are skipped for `out.ll` regeneration.
+* [litgen](chore/litgen): It generates and refreshes source-embedded `// LITTEST` FileCheck directives from the current LLVM IR for marked Go source files.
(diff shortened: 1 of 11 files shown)


zhouguangyuan0718:
Addressed the current review feedback in `cb0feeb1`:

- reused shared `// LITTEST` marker helpers from `internal/littest`
- switched module path parsing to `golang.org/x/mod/modfile`
- fixed closure anchor numbering so `init$N` counts are shared across declarations
- narrowed `genIR` panic recovery to expected panic payloads and re-panics unknown failures
- made `litgen` CLI errors user-facing instead of panicking
- switched source rewrites to temp-file + rename and tightened module-path generalization to quoted IR segments only
- added regression tests for shared init closure numbering and safer module path rewriting

Retested locally with:
- `go test ./internal/littest ./chore/gentests ./chore/litgen`
- `go test ./cl -run 'TestRunAndTestFromTestrt/litdemo$' -count=1`

@xgopilot please take another look.
@gemini-code-assist please re-review.

zhouguangyuan0718:
Addressed the latest Gemini feedback in `74862094`:

- single-file `litgen` runs now compile the containing package directory instead of the `.go` file path, so generated symbols keep the real package path instead of falling back to `command-line-arguments`
- `generalizeModulePath` now ignores backslash-escaped quotes when scanning quoted IR segments
- added regression coverage for both cases in `chore/litgen`

Retested locally with:
- `go test ./internal/littest ./chore/gentests ./chore/litgen`
- `go test ./cl -run 'TestRunAndTestFromTestrt/litdemo$' -count=1`

I did not switch global checks from `CHECK-LINE` to `CHECK`: `internal/filecheck` intentionally supports `CHECK-LINE` as a whole-line match directive, and `litgen` uses that stricter form for top-level globals.

@gemini-code-assist please take another look.

PR #1821, opened 2026-04-18
ssa: preserve nil pointer panic for large interface loads

cpunion:
This fixes a nil pointer dereference that was lost when a large dereferenced value was converted directly to an interface.

Example:

```go
type large [1 << 21]byte
var sink any
var p *large
sink = *p
```

Without this change, llgo could avoid materializing the large load by converting from the pointer path, but it also skipped the required nil pointer check and the program continued without the Go panic. The compiler now detects this large non-pointer dereference-to-interface case, emits an explicit nil check, and builds the interface from a copied value. Direct-interface ABI cases still fall back through the existing `MakeInterface` path after loading the value.

Tests added:

- `test/go`: verifies both Go and llgo observe a panic for `*p` converted to `any`.

Local verification:

- `go test ./test/go`
- `LLGO_ROOT=$PWD LLGO_BUILD_CACHE=off ./dev/llgo.sh test ./test/go`
- `LLGO_ROOT=$PWD LLGO_BUILD_CACHE=off go test ./cl ./ssa`

### DIFF
--- a/cl/builtin_test.go
+++ b/cl/builtin_test.go
@@ -58,6 +58,41 @@ func TestIsLargeNonPointerValue(t *testing.T) {
 	}
 }
 
+func TestCompileLargeNilDerefInterfaceGuards(t *testing.T) {
+	_, m := mustCompileLLPkgFromSrc(t, `
+package foo
+
+type large [1 << 21]byte
+type largeStruct struct {
+	data [1 << 21]byte
+}
+type holder struct {
+	pad [1 << 21]byte
+	value largeStruct
+}
+
+var sink any
+
+func arrayIface(p *large) {
+	sink = *p
+}
+
+func standalone(p *large) {
+	_ = *p
+}
+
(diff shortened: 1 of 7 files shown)


cpunion:
Addressed review feedback in 9751222f2:

- Removed the restrictive `*ssa.UnOp` guard for unused large dereferences and added standalone dereference coverage.
- Added base-pointer nil checks for large `FieldAddr` interface conversions.
- Replaced the hardcoded amd64/386 sizing with target pointer-size based `types.StdSizes` and named the large-deref threshold.
- Switched `MakeInterfaceFromPtr` to `Builder.AssertNilDeref`.
- Added array, struct, standalone deref, non-empty interface, and large-field tests.

The `runtime.Error` panic-value mismatch is a pre-existing runtime-wide behavior shared by the existing Assert* helpers, so I am leaving that out of scope for this focused nil-deref preservation PR.

PR #1882, opened 2026-05-22
runtime: preserve recover error values

cpunion:
## Summary
- Panic with runtime error values instead of plain strings so `recover()` values assert to `error` and `runtime.Error`.
- Build `TypeAssertionError` values for compiler-generated failed type assertions, and fix slice-to-array conversion panic bounds order.
- Preserve direct `recover` semantics through compiler-generated method wrappers when the wrapper is itself the deferred call, while keeping nested wrapper calls non-recovering.
- Preserve Go 1.26 embedded method-wrapper function pointer semantics for `fixedbugs/issue73917.go` and `fixedbugs/issue73920.go`: a recover in a real helper call reached through deferred promoted wrappers remains indirect, and the outer deferred recover still sees the panic.
- Preserve local values across fault-triggered `longjmp` recovery by emitting volatile loads/stores for local SSA allocs in functions with recover blocks.
- Add focused `test/go` coverage for recovered runtime errors, type assertion panic values, method-wrapper recover semantics, embedded wrapper function-pointer recover semantics, and fault recovery preserving named results; remove the now-passing `recover2.go`, `zerodivide.go`, `recover4.go`, `fixedbugs/issue73917.go`, and `fixedbugs/issue73920.go` GOROOT xfails.

## Still XFail / Out of Scope
- `recover.go` remains xfail in this PR. On this branch it still fails before the recover-10 path with the #1892 interface type identity issue: `interface conversion: interface {} is func(*main.T1), not func(*main.T1) (types from different scopes)`.
- This PR does not cherry-pick #1892. The recover-wrapper behavior exposed by `missing recover 10` is covered independently by `test/go`.
- The new embedded-wrapper `test/go` cases are Go 1.26+ semantic coverage. Go 1.24 has pre-Go 1.26 behavior for the upstream `issue73917.go`/`issue73920.go` programs, so those local tests skip under Go <1.26.
- A full Go 1.26 standard-library run of `test/go` currently hits unrelated nil/range behavior in `TestRangeOverNilArrayPointerCallIsEvaluated`; this PR intentionally does not cover nil/range, chan, print, interface, reflect, GC, liveness, finalizer, or goroutine domains.

## Testing
- `go test ./test/go -run 'TestRecoverThrough|TestRecoverAfterFault' -count=1`
- `go run ./cmd/llgo test -run 'TestRecoverThrough|TestRecoverAfterFault' -count=1 ./test/go`
[...]

PR #1918, opened 2026-05-24
runtime: tighten recover to direct deferred calls

cpunion:
## Summary
Fix recover/defer ownership during panic unwinding.

`recover` now only succeeds from the deferred function that is actually being run, while panics rethrown through nested defer and range-over-func frames remain recoverable by the correct outer defer.

Examples covered by this PR:

```go
// A directly deferred function value can recover the panic.
defer func() {
	if recover() == nil {
		panic("not recovered")
	}
}()
panic("boom")
```

```go
// A nested defer inside a deferred function must not recover the outer panic.
defer func() {
	defer func() {
		if recover() != nil {
			panic("nested recover should be nil")
		}
	}()
	if recover() != "outer" {
		panic("outer recover failed")
	}
}()
panic("outer")
```

```go
// A panic crossing range-over-func/defer frames remains recoverable by the
// correct deferred function, instead of being claimed by an inner frame.
defer func() {
	if recover() != "loop panic" {
		panic("recover owner mismatch")
	}
}()
for range func(yield func(int) bool) {
	defer func() { _ = recover() }()
	yield(1)
} {
	panic("loop panic")
}
```

## Main verification
After #1950 was merged, latest `xgo-dev/llgo` main still fails the nested defer/recover case above. The temporary LLGo test failed with:

```text
outer recover = <nil>, want outer
```

The same case passes with `go test`, so this PR still covers a remaining LLGo recover/defer ownership bug.


cpunion:
Added patch-coverage tests for the recover/direct-deferred-call helpers:

- `ssa/ssa_test.go`: covers recover defer token classification for marked function decls, function pointers, non-function values, nil expressions, and the recover builtin helper.
- `cl/cgo_test.go`: covers `functionUsesRecover` nil/direct-recover cases and `callMayRecover` conservative closure/default paths.

Local verification:

- `go test ./ssa -coverprofile=/tmp/llgo-recover-ssa-full-after.out` (passes; `ssa/eh.go` recover helper functions now report 100% for `callRecoverScopedDefer`, `recoverDeferToken`, `deferMayRecover`, and `isRecoverBuiltin`)
- `go test ./cl -run 'Test(RecoverCallClassificationHelpers|EmitDo|CompileRangeFuncDeferModule|DeferStackOwner|RangeFuncDefer|NestedRangeFunc|Cgo)' -coverprofile=/tmp/llgo-recover-cl-after2.out` (passes; `functionUsesRecover` and `callMayRecover` report 100%)
- `go test ./test/go -run TestRecoverFixedbug -count=1` (passes)
- `go run ./cmd/llgo test -run TestRecoverFixedbug -count=1 ./test/go/recover_defer_fixedbugs_test.go` (passes)
- `git diff --check` (passes)

Note: `go run ./cmd/llgo test -run TestRecoverFixedbug -count=1 ./test/go` still links the whole `test/go` package and failed before reaching the recover tests due unrelated generic-local-type undefined symbols, so I used the single recover test file for the focused LLGo runtime/codegen check.

zhouguangyuan0718 (approved):

cpunion:
Rebased onto #2016 (`codex/pclntab-linkphase-p1`, which includes #2012) per the review-order plan: #2012 → #2016 → this line of semantics fixes. Conflicts resolved were additive (context fields, the noinline condition set, and `runtime.Panic` now calls `SavePanicCallerFrames()` before the panic-node bookkeeping). Note the PR base is still `main`, so the diff shows #2012/#2016 commits until those merge.

cpunion:
Superseded by #2033, which re-expresses the Defer-node recover-ownership model on the stage-5 base (#2023); the funcinfo draft carried here was superseded by #2012/#2016/#2019. #1882's recover-error-value work re-expresses on top of #2033 next.

PR #1949, opened 2026-06-08
ci: retry CI archive downloads

cpunion:
Add retry handling for CI downloads that currently fail on transient GitHub 5xx responses.

ESP QEMU archives are downloaded to a temporary file with curl retries before extraction. If GitHub release assets remain unavailable, embedded emulator checks are skipped with a warning instead of failing unrelated PR jobs.

Crosscompile archive downloads now retry transient server/network errors, and GitHub archive tag URLs are fetched through codeload to avoid flaky archive redirects.

Embedded smoke checks that depend on ESP release assets now fail only for real build/test errors; external 5xx/download outages are reported as warnings.

### DIFF
--- a/.github/actions/setup-embed-deps/action.yml
+++ b/.github/actions/setup-embed-deps/action.yml
@@ -25,10 +25,16 @@ runs:
       run: |
         chmod +x .github/workflows/install-esp-qemu.sh
         QEMU_DIR=".cache/qemu"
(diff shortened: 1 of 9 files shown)


cpunion:
Closing this approach. CI download failures should fail the job directly; this PR added retry/soft-fail handling that changes that behavior.

PR #1950, opened 2026-06-08
ssa: avoid branching after panic in deferred calls

zhouguangyuan0718:
## Summary
- avoid appending an extra branch after panic-generated unreachable blocks in IfThen
- refresh recoverthenpanic and cgodefer FileCheck expectations under cl/_testgo
- cover recover-then-panic and CGo deferred call IR with valid IR checks

## Tests
- export PATH=/opt/tools/go1.26.2.linux-amd64/go/bin:$PATH; export GOROOT=/opt/tools/go1.26.2.linux-amd64/go/; go test ./cl -run '^TestRunAndTestFromTestgo$/^(cgodefer|recoverthenpanic)$' -count=1 -v
- export PATH=/opt/tools/go1.26.2.linux-amd64/go/bin:$PATH; export GOROOT=/opt/tools/go1.26.2.linux-amd64/go/; go test ./ssa -run '^TestFromTestlibgo$/^mapzero$' -count=1 -v
- export PATH=/opt/tools/go1.26.2.linux-amd64/go/bin:$PATH; export GOROOT=/opt/tools/go1.26.2.linux-amd64/go/; go test ./ssa -run 'Test.*Defer|Test.*IR' -count=1

Note: full go test ./ssa now reaches FileCheck comparisons but fails on existing unrelated _testgo fixture drift (for example abimethod), not on the previous invalid LLVM IR parse error.

### DIFF
--- a/cl/_testgo/cgodefer/cgodefer.go
+++ b/cl/_testgo/cgodefer/cgodefer.go
@@ -89,10 +89,10 @@ import "C"
 // CHECK-NEXT:   %32 = extractvalue { ptr, i64, { ptr, ptr } } %31, 0
 // CHECK-NEXT:   store ptr %32, ptr %18, align 8
 // CHECK-NEXT:   %33 = extractvalue { ptr, i64, { ptr, ptr } } %31, 2
+// CHECK-NEXT:   call void @"{{.*}}/runtime/internal/runtime.FreeDeferNode"(ptr %30)
 // CHECK-NEXT:   %34 = extractvalue { ptr, ptr } %33, 1
 // CHECK-NEXT:   %35 = extractvalue { ptr, ptr } %33, 0
 // CHECK-NEXT:   call void %35(ptr %34)
-// CHECK-NEXT:   call void @"{{.*}}/runtime/internal/runtime.FreeDeferNode"(ptr %30)
 // CHECK-NEXT:   br label %_llgo_8
 // CHECK-EMPTY:
 // CHECK-NEXT: _llgo_8:                                          ; preds = %_llgo_7, %_llgo_2
--- a/cl/_testgo/recoverthenpanic/in.go
+++ b/cl/_testgo/recoverthenpanic/in.go
@@ -27,7 +27,7 @@ package main
 // CHECK-NEXT:   %12 = getelementptr inbounds %"{{.*}}/runtime/internal/runtime.Defer", ptr %5, i32 0, i32 4
 // CHECK-NEXT:   %13 = getelementptr inbounds %"{{.*}}/runtime/internal/runtime.Defer", ptr %5, i32 0, i32 5
 // CHECK-NEXT:   store ptr null, ptr %13, align 8
-// CHECK-NEXT:   %14 = call i32 @__sigsetjmp(ptr %4, i32 0)
+// CHECK-NEXT:   %14 = call i32 @{{.*}}sigsetjmp(ptr %4, i32 0)
 // CHECK-NEXT:   %15 = icmp eq i32 %14, 0
 // CHECK-NEXT:   br i1 %15, label %_llgo_4, label %_llgo_7
 // CHECK-EMPTY:
--- a/cl/_testgo/recoverthenpanic/in.go
+++ b/cl/_testgo/recoverthenpanic/in.go
@@ -1,0 +1,186 @@
+// LITTEST
+package main
+
+// CHECK-LINE: @0 = private unnamed_addr constant [19 x i8] c"will panic in defer", align 1
+// CHECK-LINE: @1 = private unnamed_addr constant [3 x i8] c"end", align 1
+// CHECK-LINE: @2 = private unnamed_addr constant [13 x i8] c"panic in main", align 1
+
+// CHECK-LABEL: define void @"{{.*}}/cl/_testgo/recoverthenpanic.End"(){{.*}} {
+// CHECK-NEXT: _llgo_0:
+// CHECK-NEXT:   %0 = call %"{{.*}}/runtime/internal/runtime.eface" @"{{.*}}/runtime/internal/runtime.Recover"()
+// CHECK-NEXT:   %1 = call i1 @"{{.*}}/runtime/internal/runtime.EfaceEqual"(%"{{.*}}/runtime/internal/runtime.eface" %0, %"{{.*}}/runtime/internal/runtime.eface" zeroinitializer)
+// CHECK-NEXT:   %2 = xor i1 %1, true
+// CHECK-NEXT:   %3 = call ptr @"{{.*}}/runtime/internal/runtime.GetThreadDefer"()
+// CHECK-NEXT:   %4 = alloca i8, i64 {{.*}}, align 1
+// CHECK-NEXT:   %5 = call ptr @"{{.*}}/runtime/internal/runtime.AllocU"(i64 48)
+// CHECK-NEXT:   %6 = getelementptr inbounds %"{{.*}}/runtime/internal/runtime.Defer", ptr %5, i32 0, i32 0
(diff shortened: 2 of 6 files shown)


cpunion:
I think the implementation looks correct, but this PR should add regression coverage outside the existing cl/_testgo FileCheck fixtures.

Suggested coverage:

1. test/go coverage for recover-then-deferred-panic

Add a test that builds a small probe like:

- a deferred function calls recover()
- if it recovered a panic, it schedules `defer panic(recovered)`
- the same branch continues with another statement after scheduling the deferred panic

The important assertion is not only the runtime output. The pre-fix code can still compile/run some similar shapes. The test should inspect generated LLVM IR and verify that no basic block contains instructions after an `unreachable`. This catches the old bad shape where `runtime.Panic(...)` emitted `unreachable`, but `IfThen` still appended more instructions/branches in the same block.

2. test/go or ssa coverage for deferred C calls

Add a cgo probe with:

- `p := C.malloc(8)`
- `defer C.free(p)`

The IR should verify that `runtime.FreeDeferNode` is emitted before the deferred indirect C call. This covers the `callDefer` change where the defer node is released before `buildCall`.

3. ssa-level coverage for materialized deferred args

Because `FreeDeferNode` now runs before `buildCall`, add an SSA-level test showing that all deferred call data is materialized before the node is freed.

A good shape is a loop defer with a saved argument, then assert the generated IR order:

- extract the saved argument from the loaded defer node
- call `FreeDeferNode`
- call the deferred callee using the extracted argument

This documents and tests the key invariant: `buildCall` must consume only the materialized `fn,args`, and must not read the defer node after `FreeDeferNode`.

I locally verified that these checks pass on this PR head and fail on the base revision before this PR.

cpunion:
Here are the concrete test snippets I used locally. They may need minor cleanup before landing, but the important parts are the assertions and probe shapes.

```go
func llgoIRFromProbe(t *testing.T, name, src string) string {
	t.Helper()

	root := findLLGoRoot(t)
	dir, err := os.MkdirTemp(root, "."+name+"-")
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() {
		if err := os.RemoveAll(dir); err != nil {
			t.Logf("remove temp probe dir %s: %v", dir, err)
		}
	})

	mainFile := filepath.Join(dir, "main.go")
	if err := os.WriteFile(mainFile, []byte(src), 0644); err != nil {
		t.Fatal(err)
	}

	runGoCmd(t, root, "run", "./chore/llgen", filepath.ToSlash(dir))
	data, err := os.ReadFile(filepath.Join(dir, "llgo_autogen.ll"))
	if err != nil {
		t.Fatal(err)
	}
	return string(data)
}
```

Recover-then-deferred-panic IR termination coverage:

```go
const recoverThenDeferredPanicMainProbe = `package main

func end() {
	if recovered := recover(); recovered != nil {
		defer panic(recovered)
		println("will panic in defer")
	}
	println("end")
}

func main() {
	defer end()
	panic("panic in main")
}
`

func TestRecoverThenDeferredPanicIRTerminatesBlocks(t *testing.T) {
	ir := llgoIRFromProbe(t, "recover-then-deferred-panic", recoverThenDeferredPanicMainProbe)
	assertNoInstructionsAfterUnreachable(t, ir)
}

func assertNoInstructionsAfterUnreachable(t *testing.T, ir string) {
	t.Helper()

	lines := strings.Split(ir, "\n")
	for i := 0; i < len(lines); i++ {
		if strings.TrimSpace(lines[i]) != "unreachable" {
			continue
		}
		for j := i + 1; j < len(lines); j++ {
			trimmed := strings.TrimSpace(lines[j])
			if trimmed == "" {
				continue
			}
			if trimmed == "}" || isLLVMBasicBlockLabel(trimmed) {
				break
			}
			t.Fatalf("instruction after unreachable at IR line %d: %s", j+1, trimmed)
		}
	}
}

func isLLVMBasicBlockLabel(line string) bool {
	colon := strings.IndexByte(line, ':')
	if colon <= 0 {
		return false
	}
	return !strings.ContainsAny(line[:colon], " \t")
}
```

Deferred C call coverage:

```go
const cgoDeferredFreeProbe = `package main

/*
#include <stdlib.h>
*/
import "C"

func main() {
	p := C.malloc(8)
	if p == nil {
		panic("C.malloc returned nil")
	}
	defer C.free(p)
}
`

func TestCgoDeferredFreeReleasesNodeBeforeCall(t *testing.T) {
	if strings.TrimSpace(runGoCmd(t, "", "env", "CGO_ENABLED")) != "1" {
		t.Skip("cgo is disabled")
	}
	if _, err := exec.LookPath("clang"); err != nil {
		t.Skip("clang is unavailable")
	}

	ir := llgoIRFromProbe(t, "cgo-deferred-free", cgoDeferredFreeProbe)
	freeDefer := indexLineContaining(ir, "call void @", "FreeDeferNode")
	deferredCall := indexLineContaining(ir, "call void %")
	if freeDefer < 0 {
		t.Fatalf("missing FreeDeferNode call in IR:\n%s", ir)
	}
	if deferredCall < 0 {
		t.Fatalf("missing deferred indirect call in IR:\n%s", ir)
	}
	if freeDefer > deferredCall {
		t.Fatalf("FreeDeferNode must run before deferred call, got FreeDeferNode line %d after call line %d", freeDefer, deferredCall)
	}
}

func indexLineContaining(s string, parts ...string) int {
	for i, line := range strings.Split(s, "\n") {
		ok := true
		for _, part := range parts {
			if !strings.Contains(line, part) {
				ok = false
				break
			}
		}
		if ok {
			return i + 1
		}
	}
	return -1
}
```

SSA-level materialization coverage for the `buildCall` invariant:

```go
func TestDeferInLoopMaterializesArgsBeforeFree(t *testing.T) {
	prog := ssatest.NewProgram(t, nil)
	pkg := prog.NewPackage("foo", "foo")

	params := types.NewTuple(types.NewParam(token.NoPos, nil, "", types.Typ[types.Int64]))
	int64ArgSig := types.NewSignatureType(nil, nil, nil, params, nil, false)
	callee := pkg.NewFunc("callee", int64ArgSig, ssa.InGo)
	cb := callee.MakeBody(1)
	cb.Return()
	cb.EndBuild()

	fn := pkg.NewFunc("main", ssa.NoArgsNoRet, ssa.InGo)
	b := fn.MakeBody(1)
	fn.SetRecover(fn.MakeBlock())

	// Ensure entry block has a terminator like real codegen.
	b.Return()
	b.SetBlockEx(fn.Block(0), ssa.BeforeLast, true)

	arg := b.Const(constant.MakeInt64(7), prog.Int64())
	b.Defer(ssa.DeferInLoop, callee.Expr, ssa.Builder.Call, arg)
	b.EndBuild()

	ir := pkg.Module().String()
	argExtractIdx := strings.Index(ir, "extractvalue { ptr, i64, i64 }")
	freeIdx := strings.Index(ir, "FreeDeferNode")
	callIdx := strings.Index(ir, "call void @callee(i64")
	if argExtractIdx == -1 || freeIdx == -1 || callIdx == -1 {
		t.Fatalf("missing deferred arg extraction/free/call in IR, got:\n%s", ir)
	}
	if argExtractIdx > freeIdx {
		t.Fatalf("deferred arg must be materialized before FreeDeferNode, got:\n%s", ir)
	}
	if freeIdx > callIdx {
		t.Fatalf("FreeDeferNode must run before buildCall uses materialized args, got:\n%s", ir)
	}
}
```

I verified these locally as follows:

- They pass on this PR head.
- The recover IR test fails on the base revision because instructions are emitted after `unreachable`.
- The cgo/deferred-call and SSA materialization tests fail on the base revision because `FreeDeferNode` is emitted after the deferred call.

cpunion (approved):

Issue #2004, opened 2026-06-30
Proposal: runtime funcinfo metadata and fast FuncForPC lookup


cpunion:
# Proposal: Runtime Funcinfo Metadata and Fast Runtime Lookup

## Goal

Give LLGo Go-compatible runtime introspection — `runtime.Caller`, `Callers`, `CallersFrames`, `FuncForPC`, `Func.FileLine`, `runtime.Stack`, panic tracebacks — without DWARF, `llvm-symbolizer` or libunwind on the primary path, with DCE safety, controlled binary size, and performance at or beyond Go.

## Architecture (three layers + a conformance gate)

1. **Metadata model (#2012, Stage 1)** — Go-1.26-style compact function records + shared string pools + statement-level pcline records, emitted as DCE-safe sections: ELF `SHF_LINK_ORDER` (honored by `--gc-sections`), Mach-O `live_support` + one linker-private atom symbol per record (verified under ld64/lld `-dead_strip`, incl. LTO). findfunctab geometry follows Go (4KiB buckets × 16 subbuckets) with one deliberate deviation: **uint16 subbucket deltas** — LLGo has no 16-byte MINFUNC guarantee, so Go's uint8 could overflow; uint16 makes it structurally impossible. Runtime falls back to first-use table construction when the link-phase table is absent.
2. **Link-phase table (#2016, Stage 2 P1–P3)** — after linking, `internal/pclnpost` deduplicates records against the symbol table (FNV: a record is canonical iff its anchor's owning symbol hashes to its symbolID, or is that symbol's `__llgo_stub.` wrapper — LTO inline copies fail exactly), sorts, and rewrites the entry section in place with a prebuilt `ftab`+findfunctab the runtime adopts **zero-copy**. Mach-O specifics proved out: chained-fixup chain surgery, BIND-node decoding through the imports table (exported-function anchors are binds, not rebases), the header base slot spliced back as a live rebase so dyld writes the slid address, ad-hoc re-sign only when the input was signed. Overflow spills the blob into the larger stub section behind a `LLGOFTB2` redirect — rows are never dropped (a gap-y table silently mis-names function-value pcs). Startup **pre-warm** (adopt + page-touch + one synthetic lookup) mirrors Go, whose pclntab is warm by `main`.
[...]

PR #2012, opened 2026-07-01
runtime: add Go-style funcinfo find index

cpunion:
Stage 1 of #2004. Successors build on this branch: #2016 (link-phase prebuilt table), #2019 (FP unwinder), #2023 (acceptance & conformance) — each contains its predecessors until merged.

Implements the runtime funcinfo path on top of current `xgo-dev/main`, moving LLGo toward Go's `pclntab/findfunctab` model while keeping the metadata DCE-safe, plus the optimization round that followed review: Mach-O site sections, cold-path elimination, uint16 find index, Caller/Callers memoization, Frames.Next allocation removal, and debug-build gating.

## Changes

- Compact funcinfo + PC-line metadata pipeline used by `runtime.Caller`, `runtime.CallersFrames`, `runtime.FuncForPC`, `Func.FileLine`.
- `internal/pclntab`: LLVM-independent faithful port of Go's `findfunctab` algorithm (uint8 deltas, overflow error, forward scan, sentinel), kept as the reference for link-phase work.
- Go-style bucket/subbucket find index in the runtime for function-entry and PC-line tables. Deliberate deviation: **uint16 subbucket deltas** — Go guarantees a 16-byte minimum function size in its linker; LLGo indexes call-site records a few bytes apart, and a delta counts deduplicated PCs inside one 4KiB bucket (≤4096), so uint16 makes overflow structurally impossible and the index unconditional.
- **Mach-O site sections** (`__DATA,__llgo_fie/__llgo_stub/__llgo_pcl`, `live_support`, one linker-private atom symbol per record, `section$start$` boundaries with the `\x01` verbatim-name prefix): under ld64/lld `-dead_strip` a record survives exactly when its function is live — same semantics ELF gets from `SHF_LINK_ORDER` with `--gc-sections`. macOS previously had no sites at all and fell back to one dlsym per function+stub on first use (13ms cold), and had no statement-level line info.
- Cold fast path: bounded scan of raw entry/stub sections + dladdr exact-entry fallback, budgeted (8 lookups) before building the sorted table; first-use table build no longer does per-stub dlsym and materializes strings once per function/file instead of per site.
- `Caller`/`Callers` memoize interned synthetic PCs in shadow-stack slots (invalidated on line change); `Frames.Next` reuses the FuncForPC 4-way cache instead of allocating a `*Func` per frame.
- Debug builds (`LLGO_DEBUG`/`LLGO_DEBUG_SYMBOLS`) keep the funcinfo tables but drop the body-embedded site records — the in-body asm anchors shift instruction/scope layout enough to confuse debuggers (LLDB suite now 194/194). `runtime.Caller` is unaffected (instrumentation is independent); `FuncForPC(entry)`/`FileLine(entry)` work via the dlsym fallback; statement/call-site line granularity is disabled in debug builds.
- Observability: `LLGO_FUNCINFO_DEBUG=1` prints per-table `frames/buckets/index=built|fallback/entries=sites|dlsym` so benchmarks can prove which path they measured. `LLGO_FUNCINFO_SITES=0` keeps tables but drops all site asm (see Known limitations).

## Benchmarks (final HEAD, 5–7 runs, best/trimmed avg)

macOS arm64 (host) and Linux arm64 (container); `main` = merge base; 576 target functions (24 pkgs × 24 methods). Index proof: `LLGO_FUNCINFO_DEBUG=1` reports `index=built` for func and pcline tables in all measured scenarios.

| metric | mac go1.26 | mac this PR | mac +LTO | linux go1.26 | linux this PR | linux +LTO |
|---|---:|---:|---:|---:|---:|---:|
| hot.Caller0 | 148ns | **56ns** | **44ns** | 161ns | **67ns** | **57ns** |
| hot.Caller1 | 166ns | **86ns** | **67ns** | 174ns | **109ns** | **94ns** |
| hot.CallersOnly | 124ns | 130ns | **101ns** | 123ns | 156ns | 131ns |
| hot.CallersFramesFirst | 419ns | 452ns | **341ns** | 437ns | 533ns | **444ns** |
| hot.FuncForPCEntry | 12ns | **9ns** | **6ns** | 12ns | **9ns** | **7ns** |
| multipkg.FuncForPCMany (576) | 9.9µs | **3.8µs** | **879ns** | 10.6µs | **4.2µs** | **893ns** |
| multipkg.FileLineMany (576) | 31.4µs | **5.2µs** | **1.3µs** | 31.8µs | **5.4µs** | **1.4µs** |
| cold.FirstCaller0 | 2.4µs | **666ns** | 458ns | 3.5µs | **1.7µs** | 2.1µs |
| cold.FirstFileLine | 167ns | **167ns** | **42ns** | 333ns | **83ns** | 125ns |
| cold first-use table build¹ | — | ~45µs | ~25µs | — | ~55µs | ~55µs |
| stdlib.Work² | 5.4µs | 17.7µs | 14.6µs | 6.2µs | 20.3µs | 17.6µs |

¹ One-time per process; shows up in whichever `cold.First*` window exhausts the cold-lookup budget (platform-dependent). Go needs none because its linker ships the sorted table — this is the link-phase follow-up.
² Shadow-stack instrumentation tax; the `main` baseline is equally behind Go (18–20µs), not introduced by this PR — unwinder follow-up.

Scope of the comparison: `FileLineMany` compares LLGo's statement/call-site granularity against Go's dense per-instruction pcvalue tables; "faster than Go" holds for this granularity, not for full pcvalue semantics.

Scaling: per-target batch cost is flat from 144 → 576 targets (6.5 → 6.8ns per FuncForPC target, LTO ~1.9ns), consistent with the bucket index.

## Binary size and build time (`main` → this PR)

| scenario | mac | mac LTO | linux | linux LTO | vs go1.26 |
|---|---|---|---|---|---|
| multipkg (576) | 2.4→3.1 MiB (+29%) | 1.8→2.7 (+50%) | 2.4→3.3 (+38%) | 2.2→3.6 (+64%) | above (2.7–2.8) |
| stdlib | 3.4→4.0 MiB (+18%) | 2.7→3.5 (+30%) | 3.3→4.0 (+21%) | 3.1→4.5 (+45%) | below (5.0–5.1) |

Composition (stdlib, +660KB): strings ~270KB (content-deduped, comparable density to Go's funcnametab), stub sites 115KB, entry sites 70KB, records+symbol-index+hash ~156KB, pcline <5KB. Anchor: the same program's Go `__gopclntab` is 1.41MB — this PR's total metadata is ~47% of Go's equivalent structure. Build time +2–5% (non-LTO), +8–17% (LTO).

## Impact on ordinary code (no runtime introspection)

Pure-compute benchmark (json/sort/map/fib), interleaved runs. Narrow A/B via `LLGO_FUNCINFO_SITES=0` (tables kept, only in-body site asm removed):

| metric | main | sites-off | sites-on | attribution |
|---|---:|---:|---:|---|
| fib30 | 2307µs | 2305µs | 1624µs (−29.5%) | site asm shifts inline/layout decisions |
| json | 47.9ms | 48.8ms | 47.8ms (−2.1%) | noise |
| sort | 120.2ms | 121.4ms | 121.1ms (−0.3%) | noise |
| map | 104.5ms | 102.7ms | 108.5ms (+5.7%) | site asm in runtime-internal hot funcs |

`sites-off ≈ main` within noise on every metric: the metadata tables have zero runtime impact; all perturbation comes from the body-embedded volatile asm, in both directions. Go does not have this problem — its pclntab is generated by the linker and never enters the optimizer. Against Go, ordinary-code performance is dominated by the LLGo baseline (json 2.3×, sort 4.9×, map 3.1× slower on both `main` and this PR; fib 39% faster) — funcinfo's ±5% is second-order there.

## Known limitations / follow-ups

1. **First-use table construction (~45–55µs once per process)** and ~35% of the size delta (site sections + symbol index + hash) are removed together by link-phase `ftab`/`findfunctab` generation — design doc in `doc/design/pclntab-linkphase.md`, `internal/pclntab` is the ready algorithm base.
2. **LTO inlining duplicates body-embedded entry records** (`frames 5726 → 15106` at 576 targets; ELF section 4.1×), registering host-function PCs under the inlinee's symbol ID; runtime only consults the table when native symbolization fails, which bounds the impact. IR-level fixes were tried and ruled out (`!associated` is linker-GC-only and GlobalDCE deletes the records; `llvm.compiler.used` pins dead functions; `noduplicate` blocks inlining) — dedup is link-phase work; `!pcsections` metadata is the candidate replacement for the asm mechanism (also removes the codegen perturbation above).
3. **Shadow-stack instrumentation tax** (stdlib.Work ~3×, deep-stack CallersFrames 1.5–1.8×, non-LTO CallersOnly gap on Linux) — frame-pointer unwinder follow-up.
4. Observed: `main`+LTO binaries fail at runtime on macOS in this harness; this PR's LTO passes all scenarios (root cause of the baseline failure not investigated here).

## Local validation

macOS arm64 + Linux arm64 container: `go test ./cl` (full), `go test ./internal/build ./internal/pclntab ./ssa`, `LLGO_ROOT=... go test ./test/go -run TestRuntime...`, `bash _lldb/runtest.sh` (194/194), `go fmt`/`go vet` clean. Benchmark harness included under `benchmark/runtime_funcinfo` (reproduces all numbers above).


cpunion:
Folding into #2024: the stage-5 backbone (#2012 -> #2016 -> #2019 -> #2023) lands as one PR through #2024, which contains all of these commits rebased onto current main.

PR #2016, opened 2026-07-02
runtime: link-phase ftab/findfunctab generation (Stage 2, P1–P3)

cpunion:
Stage 2 of #2004, based on #2012 (contains it until merged); successors: #2019 (FP unwinder), #2023 (acceptance & conformance).

Follow-up to #2012: link-phase `ftab`/`findfunctab` generation, per `doc/design/pclntab-linkphase.md` (in this branch). **P1 (analysis), P2 (automatic write-back + zero-copy runtime adoption) and P3 (Mach-O bind-record resolution, startup pre-warm, scale hardening) are complete**; every `llgo`-linked executable on linux/darwin now ships a prebuilt function table that is adopted zero-copy and pre-warmed at startup.

## How it works

- After `linkMainPkg`, `internal/build` runs `internal/pclnpost.Rewrite` on the executable: parse the funcinfo entry/stub site sections, identify canonical records by FNV (a record survives when its anchor's owning symbol hashes to its symbolID, or is that symbol's `__llgo_stub.` wrapper — LTO inline copies fail this test exactly), sort, build the findfunctab, and rewrite the entry section in place (header + `{entryOff, funcIndex}` ftab + buckets), voiding the merged stub section. Any failure degrades silently to #2012's first-use construction (`LLGO_PCLNPOST=0` forces that path).
- The runtime adopts the table **zero-copy** when the magic header validates: lookups binary-search the on-disk ftab through the shared bucket index; nothing is materialized on first use. Entries are true symbol starts, retiring the entry-PC slack on this path.
- **Mach-O bind records (P3)**: pointer slots that name *exported* functions — every `__llgo_stub.*` wrapper and any exported Go function — are chained-fixup **BIND** nodes, not rebases, even though they bind back into the same image. The rewriter resolves them through the `LC_DYLD_CHAINED_FIXUPS` imports table. Before this, all stub records were dropped as unowned, so `FuncForPC` on a function value silently fell back to **dladdr (~6µs per fresh pc)** on darwin — the real cause of the "mac cold gap" previously attributed to re-signed-binary page validation.
- The prebuilt header's base slot is spliced back into the fixup chain as a live rebase node: dyld writes the *slid* text base at load, and the runtime reads a ready runtime PC with no slide arithmetic (on non-PIE ELF the link-time value already equals the runtime address). An insert that joins no chain page fails the rewrite loudly → first-use fallback.
[...]

PR #2019, opened 2026-07-03
runtime: LLGo-owned frame-pointer unwinder (Stage 5)

cpunion:
Stage 5 of #2004, based on #2016 (contains #2012+#2016 until merged); successor: #2023 (end-user acceptance & conformance). Replaces the shadow-stack instrumentation and libunwind with an LLGo-owned frame-pointer unwinder.

## What changes

- **Every Go function keeps the frame-pointer chain** (`"frame-pointer"="non-leaf"`, verified `stp x29,x30 / mov x29,sp` prologues). Cost is the standard ~1-2% envelope — Go itself always keeps FP on arm64.
- **`runtime.Caller/Callers/CallersFrames/Stack` walk real stacks** (`fpCallers`: [fp]/[fp+8] chain with stride/alignment guards), symbolized through #2016's prebuilt ftab + pcline labels. Return addresses resolve at `pc-1` (Go's convention); the machine scheduler can place the next statement's label exactly on a return address, so raw-pc lookups mis-attributed frames.
- **Shadow-stack instrumentation is no longer emitted** (`LLGO_SHADOW_STACK=1` keeps the legacy emitters for one release). Tracked functions retain only `noinline`, no-tail-call and the data-only pcline records.
- **libunwind is gone**: the clite stacktrace (unrecovered-panic dump, last-resort Callers fallback) walks the FP chain with dladdr names — same output format, no `-lunwind`.
- **Explicit FP pairing flag**: the compiler emits a per-binary `__llgo_fp_chain` byte next to the funcinfo table recording `ssa.Program.NeedsFramePointer()`; `fpUnwindAvailable` trusts that declaration plus table presence, so a target that keeps tables but not the FP attribute can never take the physical walk by accident.
- **One symbolization rule**: every path that resolves a function record refines it through a single `refinePCSymbolLine` helper (same-function statement record at pc, then pc-1) — `FuncForPC`, `FileLine` and `CallersFrames` cannot disagree on line attribution.
- Collector fixes along the way: methods and anonymous functions were never trackable (method frames had no pcline labels; closures lost their innermost frame to tail-call optimization). #2023 extends the collector further (directly-called methods, uniform stdlib tracking).

## Semantics: now Go-conformant, verified against gc

Physical stacks see every real frame — the shadow stack only ever recorded instrumented ones. Skip counts in the statement-line probes moved to gc ground truth (each verified by running the same chain shapes under go): interface chain MARK at skip 3 (was 2), closure chain at skip 4 (was 3). Adjacent `runtime.Stack` calls report their own lines. Mid-body pcs, `FuncForPC(pc-1)` and `FileLine(pc-1)` agree with `Frames`.

## Performance

All numbers `best/trimmed-avg`, 5 runs (scale rows 3), 24×24=576-target scenarios, ±LTO. `s5` = this PR.

### Hot paths — macOS / Linux (ns)

| metric | go1.26 | 2016 | **s5** | **s5+lto** |
|---|---:|---:|---:|---:|
| hot.Caller0 | 155 / 241 | 55 / 66 | **17 / 37** | **14 / 29** |
| hot.Caller1 | 173 / 246 | 86 / 111 | **17 / 39** | **14 / 31** |
| hot.CallersOnly | 129 / 174 | 131 / 153 | **31 / 56** | **28 / 47** |
| hot.CallersFramesFirst | 270 / 626 | 455 / 544 | **145 / 554** | **121 / 406** |
| hot.FuncForPCEntry | 13 / 17 | 9 / 9 | **2 / 2** | **1 / 1** |
| hot.FuncFileLineEntry | 10 / 14 | 10 / 11 | **3 / 4** | **1 / 2** |
| stdlib.Caller0 | 146 / 141 | 75 / 109 | **15 / 26** | **12 / 21** |

### Deep stacks — the Stage 5 headline (macOS / Linux)

| metric | go1.26 | 2016 | **s5** |
|---|---:|---:|---:|
| deep.Direct32 | 791ns / 937ns | 1.3µs / 1.7µs | **44ns / 115ns** |
| deep.CallersFramesAll (32) | 2.1µs / 2.4µs | 3.2µs / 4.0µs | **423ns / ~1µs** |
| deep.Direct512 | 9.7µs / 9.8µs | 87µs / 95µs | **2.8µs / 2.9µs** |
| deep.CallersFramesAll (512) | 11.5µs / 12.2µs | 105µs / 111µs | **3.7µs / 11.3µs** |

The ~160ns/frame shadow-stack tax is gone: 512-deep chains with a `Caller` at the bottom are now **3.4× faster than Go**, and instrumented-graph big bodies dropped 24×:

| bigfunc 16×2000 (macOS) | go1.26 | 2016 | **s5** | **s5+lto** |
|---|---:|---:|---:|---:|
| bigfunc.Work | 30.0µs | 433µs | **18.1µs** | **17.6µs** |
| bigfunc.FuncForPCMid | 63ns | 149ns | **27ns** | **16ns** |
| bigfunc.FileLineMid | 86µs | 176ns | **57ns** | **27ns** |
| bigfunc.CallersFramesMid | 1.0µs | 329ns | 310ns | **186ns** |

### Cold paths (first use per process, macOS / Linux)

| metric | go1.26 | 2016 | **s5** |
|---|---:|---:|---:|
| cold.FirstCaller0 | 3.7µs / 3.2µs | 500ns / 667ns | 584ns / 1.3µs |
| cold.FirstFuncForPC | 2.8µs / 1.2µs | 2.3µs / 2.8µs | **2.5µs / 1.7µs** |
| cold.FirstCallersFrames | 2.8µs / 1.2µs | 1.3µs / 3.1µs | 4.6µs / 4.0µs |
| cold.FirstFileLine | 834ns / 292ns | 83ns / 166ns | **41ns / 41ns** |

### Ordinary code and size

`plain.*` (fib/json/sort/map) is identical across 2012/2016/s5 within noise on both platforms — code that never asks for caller info pays nothing. Binary sizes are unchanged (±0.1MiB across every scenario); linux binaries no longer link `-lunwind`.

### Known follow-ups (single cells, documented)

- `bigfunc.FirstFileLineMid` (32k-site probe): 18.5µs vs 2016's 1.0µs — first-use interaction between the startup pcline warm and site-dense binaries; P4's prebuilt pcline table subsumes it.
- `stdlib.Work` is unchanged (~23µs vs Go 6.2µs): the residual is `noinline`/no-tail-call on Caller-reachable functions plus the LLGo baseline, no longer unwind bookkeeping. Lifting noinline needs the P4 inline tree.
- `mac cold.FirstCallersFrames` 4.6µs vs 2016's 1.3µs: first-walk memoization allocations; steady-state walks are 5–8× faster than 2016.

## Validation

- macOS: `cl` (462s), `test/go`, `internal/build`, `ssa`, LLDB suite **0 failed** — all with shadow-stack emission off.
- Linux (container): `test/go` green with `-lunwind` removed.
- IR goldens updated (`attributes #0` now carries the FP attribute; out.ll files carry no attribute groups and needed no regeneration).

## Known scope cuts (documented)

- Inlined frames: uninstrumented functions may still be inlined by LLVM and their frames elided (same as C); tracked functions keep `noinline`. The P4 inline tree lifts this.
- Foreign frames: the walk stops at the first frame that breaks chain discipline (C code without frame pointers, e.g. x86-64 libs built with omission). arm64 C code keeps FP by default.

Depends on #2016 (branched from it; rebases after it merges). #2023 builds on this PR and carries the end-user acceptance suite.


cpunion:
Folding into #2024 (contains this PR's commits rebased onto current main; see #2012 for the chain note).

PR #2022, opened 2026-07-03
runtime,cl,build: stage5 cleanup — explicit FP pairing flag and tidy-ups

cpunion:
Post-review cleanup for the Stage 5 unwinder (#2019). No behavior change on supported targets; one hardening change in how the runtime decides FP-walk availability.

## Changes

- **Explicit FP pairing flag**: the compiler now emits a per-binary `__llgo_fp_chain` byte next to the funcinfo table (`internal/build.emitFuncInfoTable`), recording `ssa.Program.NeedsFramePointer()`. `fpUnwindAvailable` checks the flag plus table presence instead of inferring the chain from table presence alone — a binary whose target keeps tables but not the FP attribute can never take the physical walk by accident.
- **`callerLocation` helper**: the `"???"`/line-1 normalization in `runtime.Caller` was duplicated three times.
- **Named constants**: `minLegalPC` (zero page — never code), `maxFPStride` (corrupt-chain bound), `frameSymbolResultCacheSize`; previously three unrelated meanings shared the literal 4096 and two shared `1<<20`.
- **Walker cross-references**: `fpCallers` (Go) and `llgo_stacktrace` (debug.c) implement the same chain discipline; each now points at the other so they stay in sync.
- **Dead weight**: removed the `LLGO_DBG_PCLINE` per-label debug print (a `Getenv` per emitted label) and a duplicate `unsafe` import.
- **Docs**: `doc/design/stage5-handoff.md` trimmed from the session-by-session investigation log to the durable invariants (empty-prebuilt fallback, amd64 pcline merging, walker sync, target gating) and diagnostic traps.

Deliberately left for the shadow-stack removal (one release out): merging the two symbol caches, simplifying `funcPCFrameForEntryPC`, deleting the legacy emitters.

## Validation

- `go build ./...`, `go vet` (known issues only), funcinfo table tests incl. new `__llgo_fp_chain` assertions (both values).
- End-to-end `test/go` runtime line/stack/caller tests green with the flag wired through.

Based on #2019 (contains #2012 and #2016); rebases as those merge.


cpunion:
Folded into #2019: the cleanup touched only code that PR introduces, so reviewers now see the final form in one place (identical tree, both platforms already verified on this branch).

PR #2023, opened 2026-07-03
runtime,cl: Go-style panic tracebacks and exact log/slog/testing locations, with gc-verified acceptance suite

cpunion:
End-user acceptance for the Stage 5 unwinder (#2019): make the caller info an application developer actually sees correct — panic tracebacks, log/slog locations, testing failure lines — and lock it in with a gc-verified acceptance suite.

## What the acceptance run exposed (before this PR)

| Scenario | gc | llgo before | llgo after |
|---|---|---|---|
| unrecovered panic | goroutine trace with file:line per frame | `[0x… main+0x4, SP = 0x24]` (dladdr, one frame, no file:line) | Go-style traceback |
| `log.Lshortfile` | `main.go:11:` | `???:1:` | `main.go:11:` |
| slog `AddSource` | `source=…/main.go:12` | `source=:0` | `source=…/main.go:12` |
| `t.Errorf` under `llgo test` | `x_test.go:6:` | `???:1:` | `x_test.go:6:` |

## Fixes

- **Go-style panic tracebacks**: unrecovered panics print `goroutine 1 [running]:` plus `function(...)` / indented `file:line` per physical frame, via a `PanicTraceback` hook the public runtime registers with the runtime core; the clite dladdr dump remains the fallback when the FP walk or tables are unavailable. Runtime-internal plumbing frames are filtered by package path (gc hides runtime frames the same way).
- **Uniform caller-frame tracking — no package whitelist**: the blanket stdlib exclusion is gone. The existing per-package "transitively reaches runtime.Caller" analysis now applies to stdlib exactly as it always did to third-party code, so `log.Output`, slog's `Logger.log` and testing's decorate chains keep physical frames and fixed Caller depths count correctly. Packages that never read caller pcs still track nothing and pay nothing.
- **Cross-package call-site anchors**: a call into a caller-pc-consuming function of another package (the frame a fixed depth attributes) gets a statement anchor, so `log.Println("x")` reports the exact line even though the calling function itself never reads caller pcs. Decided by the same analysis (memoized per package), not by name.
- **Collector: directly-called methods**: a type used only concretely never enters `RuntimeTypes`, so its methods escaped tracking — exactly how `slog.(*Logger).Info` lost its frame. Methods are now collected from the package's own type declarations (both receiver forms).

- **gc-conformant naming and format**: the main package's display name is normalized to `main` regardless of the module name (frame filters in the wild match on the `main.` prefix; linker symbols keep the package path), closures display as gc's `.funcN` instead of `$N`, and traceback lines carry gc's ` +0x<pc-entry>` offset (value is codegen-specific, format matches).
- **Program-unique frames pinned**: `main.main` and package `init` run once, so noinline is free — they are the bottom frames of nearly every panic traceback and now report exact statement lines. `//go:noinline` functions already keep their frames and get statement anchors too.

Conformance bar (agreed): format and user-code file:line match gc exactly; runtime-internal, patched-stdlib and startup frames may differ (different runtime, accepted).

Cost: hello-world grows 368 bytes (the traceback printer); no IR golden changed; `plain.*` code paths are untouched.

## Acceptance suite (`test/go/caller_acceptance_test.go`)

Four probes, every expectation verified against gc: panic traceback contents; `log.Lshortfile` + slog `AddSource` (text and JSON handlers, package functions and logger methods); a failing `t.Errorf` under `llgo test`; and an introspection grab-bag — goroutine/init/defer callers, `FuncForPC` names for value/pointer methods, closures and generic instantiations, and the errors-with-stack capture idiom (zap/sentry style).

Documented divergences observed while writing the suite:
- Closure display names are normalized to gc's `.funcN` form (linker symbols keep `$N`); the suite accepts both forms for robustness.
- Linker ICF folds identical function bodies to one address, so `FuncForPC` reports the surviving symbol (gc does not fold); probes use distinct bodies.
- Cross-package wrapper chains (zap→zapcore style) are outside the per-package analysis; their wrappers may still inline. The P4 inline tree subsumes this.
- Plain (non-noinline, non-pc-consuming) user functions appearing mid-traceback in a panic still show declaration-adjacent lines: statement anchors structurally require the containing function to survive (the ELF pcline section is link-ordered to its symbol), so full coverage without noinline needs the P4 metadata mechanisms. Bottom frames (`main.main`, `init`), `//go:noinline` functions, and every attributed frame in the log/slog/testing/Caller scenarios are exact.

## Manual playground (`test/_manualtest/`, committed)

Five directories runnable with both `go` and `llgo` for side-by-side comparison: `panic/` (traceback shape), `logging/` (log/slog locations), `callers/` (Caller ladder, frames, FuncForPC names, errors-with-stack), `testfail/` (`llgo test` failure lines), `cexcept/` (hardware faults in C called from Go — segv recovery verified on darwin/arm64 + linux/arm64/amd64; known gaps recorded in its README: fault-site stack needs the panic-snapshot follow-up, SIGFPE/SIGBUS handlers and sigaltstack not installed yet). Every scenario also has a `test/go` regression: the acceptance tests above plus `TestCallerAcceptanceModuleMainNaming` (module-named main package must still report `main.*`) and `TestCallerAcceptanceCFaultRecover` (C NULL-store converts to a recoverable Go panic).

## Validation (both platforms, before every push)

- Acceptance suite 4/4 on macOS arm64 and linux arm64 (container); statement-line and line-info suites still green.
- `cl` (725s), `ssa`, `internal/build`, `test/go` all ok; `gentests` produced no golden diffs.
- Linux container: every `_demo/go` demo passes plain and an LTO (`-lto=full -globaldce`) subset — including the reflect demos whose ELF pcline sections dangled in the first CI round (anchors in inlinable functions; fixed by making frame-keeping a tracking criterion rather than a per-call exception).
- Panic traceback offset verified against disassembly (`+0x30` = return address − function entry for the probe).

Based on #2019 (contains #2012 and #2016); rebases as those merge.


cpunion:
Folding into #2024 (contains this PR's commits rebased onto current main; see #2012 for the chain note).

PR #2027, opened 2026-07-04
runtime,cl: add Go-compatible sampled memory profiling

### LINKED ISSUE #2351: Proposal: low-overhead stack-keyed sampled heap profiling
## Summary

Implement Go-compatible sampled heap profiles on native LLGo targets while keeping profiling work out of ordinary executables and minimizing the enabled allocator fast path.

There are two target paths:

- native targets with physical stack walking use stack-and-size keyed samples;
- wasm and bare-metal targets retain the existing size-class fallback until they have an equivalent stack source.

Allocator replacement and sampled-object death tracking are separate work because they depend on allocator and GC lifecycle details.

## Goals

- preserve the public `runtime.MemProfile`, `runtime.MemProfileRate`, and `runtime/pprof` contracts;
- attribute native samples to user allocation stacks and source lines;
- use randomized exponential thresholds and expose raw samples for Go's normal Poisson correction;
- give ordinary executable builds no profiler work in `AllocZ` or `AllocU` when profiling is provably unused;
- keep the enabled common path to a rate load, one TLS lookup, and a countdown update;
- avoid global lock contention for existing buckets and snapshots;
- preserve correctness under concurrent allocation and profile collection;
- make native, wasm, bare-metal, LTO, and library behavior explicit.

### 1. Native and fallback recorders

Native LLGo builds that retain profiling install a physical-stack capture hook during runtime initialization; proven non-consumer executables omit this setup entirely. Their allocator recorder has no fallback-mode dispatch. Wasm, bare-metal, and host-tool builds compile a separate size-class recorder.

A native allocation checks `MemProfileRate` before touching TLS. Rate zero returns without a TLS access. Otherwise, one physical-thread-local state holds the sampling countdown and RNG state. Threshold generation, recursion suppression, stack walking, bucket creation, and random-number work remain in a noinline sampled slow path. This matches LLGo's current 1:1 native-thread execution model and avoids shared countdown races.

### 2. Whole-program disabled-path selection

After all Go SSA is available but before package backends start, the build coordinator conservatively detects whether non-runtime code consumes `runtime.MemProfile`, `runtime/pprof`, or `runtime.MemProfileRate`.

- For a proven non-consumer executable, code generation omits the direct `recordMemProfileAlloc` calls from the internal runtime allocator and the public runtime profile-setup call. Frame-table initialization, hook installation, capture, and recorder bodies then become unreachable. `AllocZ` and `AllocU` keep their existing names, signatures, and ABI; there are no alternate allocator entries or runtime mode branches.
- Any uncertain executable keeps recording. False positives cost performance but never correctness.
- `c-archive` and `c-shared` always keep recording because future external calls are not visible during whole-program analysis.
- Full-LTO and non-LTO builds use the same pre-backend decision. Linker dead stripping then removes the unused profiler implementation from non-consumer executables.

The decision is copied to isolated backend programs and included only in the internal runtime package's cache fingerprint. Switching between profiling and non-profiling executables cannot reuse an incompatible allocator archive and does not make the setting part of user-package ABI.

### 3. Bucket key and concurrency

A native bucket is keyed by `(stack, allocation size)`, matching Go's correction model for mixed-size allocations at one call site. Bucket nodes and `next` links are immutable after publication.

Lookup traverses an atomically loaded head without a lock. On a miss, the slow path allocates a node, takes the insertion lock, rechecks, initializes counters, and atomically publishes it. Existing counters use atomic updates. A snapshot captures each table head once and traverses immutable chains without stopping allocators.

Profile materialization temporarily suppresses sampling on its physical thread. This prevents rate-1 profiling from recursively profiling symbolization and buffer growth while leaving other threads active.

### 4. Stack attribution

Allocator and capture plumbing is trimmed from exposed records. Profile-enabled compilation preserves physical frames on compiler-classified allocation paths across the whole static call graph, including cross-package wrappers, rather than disabling inlining for unrelated helpers. A general inline-call-site representation remains the preferred future replacement for retaining physical frames.

### 5. Compatibility

Changing `MemProfileRate` at runtime remains supported in builds that retain profiling. Rate `1` samples every allocation; rate `0` disables recording immediately. Default rates store raw counts so existing Go consumers apply their normal correction.

Wasm and bare-metal keep the size-class fallback. Moving either target to stack-keyed recording later does not require a Go API or allocator ABI change.

## Validation

Required coverage includes:

- native Darwin/arm64 and Linux/amd64 stack attribution;
- rate `0`, default rate, rate `1`, and runtime rate changes;
- same-stack mixed-size allocations producing distinct buckets;
- concurrent native sampling and snapshots;
- `runtime/pprof` heap output and Go GOROOT `heapsampling.go`;
- wasm and bare-metal fallback compilation;
- executable, `c-archive`, `c-shared`, non-LTO, and full-LTO selection;
- final-symbol and disassembly checks proving a non-consumer executable has no allocator hook;
- allocation microbenchmarks separating allocator cost from disabled-hook, enabled-countdown, and sampled slow-path cost.

## Prototype results

PR #2027 follows the design above without an allocator ABI or target-policy change. On native non-consumer executables, `recordMemProfileAlloc`, stack capture, and bucket symbols are absent, and `AllocZ` contains only allocation and zeroing. The enabled path resolves TLS once and enters the noinline slow path only at a sampling boundary.

On an Apple M4 Max (Darwin/arm64, Go 1.26.5), a GC-disabled allocation microbenchmark used 21 rotated rounds; each process ran 5 warm-up and 61 measured batches of 250,000 escaping 16-byte allocations. The table reports medians of process medians; signed differences subtract those displayed medians, while percentages are paired within each round.

| Toolchain and path | ns/op | Signed difference / paired change |
| --- | ---: | ---: |
| LLGo PR base (`0f480253`): unconditional size-class recorder | 19.11 | baseline |
| PR #2027: no profile consumer | 18.04 | -1.07 ns/op / -6.5% vs PR base |
| PR #2027: consumer, rate 0 | 18.99 | +0.95 ns/op / +6.1% vs omitted; -0.12 ns/op / -2.4% vs PR base |
| PR #2027: consumer, default rate | 20.03 | +1.99 ns/op / +11.1% vs omitted; +0.92 ns/op / +5.8% vs PR base |
| Go 1.26.5: rate 0 | 12.69 | Go baseline |
| Go 1.26.5: default rate | 12.84 | +0.15 ns/op / +0.3% vs Go rate 0 |

The standard Go toolchain has no no-consumer specialization, so rate 0 is its closest disabled path. Against Go's corresponding rate-0/rate-0/default paths, LLGo's omitted/rate-0/default medians differ by +5.35/+6.30/+7.19 ns/op and the paired changes are +39.9%/+48.7%/+53.0%. This gap includes allocator/runtime differences; profiling overhead is represented by the within-toolchain comparisons. LLGo used `GC_disable`; Go used `debug.SetGCPercent(-1)`.

## Expected benefits

- no profiling calls, eager setup, or reachable profiler bodies in ordinary executables where profiling is unused;
- no per-allocation function call, repeated TLS resolution, RNG work, or fallback dispatch in profiled native programs;
- no global serialization for repeated samples or snapshots;
- correct Poisson correction for mixed allocation sizes;
- unchanged public allocator ABI and isolated target fallbacks.

Minimal Linux `cprintf` remains +256 B on disk because 16 additional pre-DCE funcinfo symbol-index records (16 B each) remain in `.rodata`; `.relro_padding` shrinks by the same amount, so `.text`, loaded-section total, and page count are unchanged. This is a metadata-filtering limitation, not retained profiler machine code.

## Non-goals and rollout

PR #2027 is the implementation vehicle for stack-keyed sampling, the native hot-path split, whole-program hook omission, correct bucket keys, lock-free lookup/snapshot, attribution narrowing, and selection tests. Allocator-specific tuning, sampled-object free accounting, improved wasm/bare-metal stack capture, and dead funcinfo filtering remain follow-up work.

cpunion:
Closes #2351.

## Problem

LLGo's `runtime.MemProfile` only kept size-class counters, so native heap profiles lacked allocation stacks and source-line attribution. The initial stack-sampling implementation also put avoidable TLS, locking, buffer-allocation, and frame-retention costs on profiled programs.

## Design

- Native targets sample with randomized exponential per-thread thresholds and key raw samples by `(physical stack, allocation size)`, matching Go's Poisson-correction model.
- Existing buckets and snapshots use atomically published immutable chains; only first insertion takes the table lock. Profile materialization suppresses recursive sampling on its current physical thread.
- `runtime.MemProfile` reuses the caller's compatible record buffer, and `runtime/pprof` keeps sampling paused through its buffer allocation and snapshot materialization.
- After whole-program Go SSA construction, an executable that does not consume `runtime.MemProfile`, `runtime.MemProfileRate`, or `runtime/pprof` omits both allocator recording calls and native profile setup at code generation. Frame-table initialization, hook installation, capture, and recorder bodies are then unreachable and dead-stripped; allocator names, signatures, and ABI do not change.
- Profile-enabled builds retain physical frames only along compiler-classified allocation paths across the static call graph, including cross-package wrappers. Unrelated helpers remain inlineable.
- Native TLS addresses are dominance-cached, so the enabled allocator path resolves TLS once. Random generation and stack walking remain in a noinline sampled slow path.
- Wasm and bare-metal retain the existing size-class implementation until equivalent stack capture exists. `c-archive` and `c-shared` conservatively retain profiling because future external calls are not visible at executable analysis time.
- The whole-program choice is stored only in the internal runtime cache fingerprint and is shared by normal and full-LTO builds.

| Output | Recording path |
| --- | --- |
| Native executable, no consumer | allocator recording and profile setup omitted; profiler dead-stripped |
| Native executable with consumer | stack-and-size sampled profile |
| Wasm/bare-metal executable with consumer | existing size-class fallback |
| `c-archive` / `c-shared` | profiling retained; target-appropriate recorder |

## Benefits

- Ordinary executables have no profiling calls or eager profile setup; profiler bodies are dead-stripped.
- The enabled common path performs one rate load, one TLS resolution, and one countdown update; it has no RNG, stack walk, global lock, or target-mode dispatch.
- Mixed-size allocations at one stack are corrected independently, and concurrent snapshots no longer serialize allocation samples.

On an Apple M4 Max (Darwin/arm64, Go 1.26.5), a GC-disabled allocation microbenchmark used 21 rotated rounds; each process ran 5 warm-up and 61 measured batches of 250,000 escaping 16-byte allocations. The table reports medians of process medians; signed differences subtract those displayed medians, while percentages are paired within each round.

| Toolchain and path | ns/op | Signed difference / paired change |
| --- | ---: | ---: |
| LLGo PR base (`0f480253`): unconditional size-class recorder | 19.11 | baseline |
| This PR: no profile consumer | 18.04 | -1.07 ns/op / -6.5% vs PR base |
| This PR: consumer, rate 0 | 18.99 | +0.95 ns/op / +6.1% vs omitted; -0.12 ns/op / -2.4% vs PR base |
| This PR: consumer, default rate | 20.03 | +1.99 ns/op / +11.1% vs omitted; +0.92 ns/op / +5.8% vs PR base |
| Go 1.26.5: rate 0 | 12.69 | Go baseline |
| Go 1.26.5: default rate | 12.84 | +0.15 ns/op / +0.3% vs Go rate 0 |

The standard Go toolchain has no no-consumer specialization, so rate 0 is its closest disabled path. Against Go's corresponding rate-0/rate-0/default paths, LLGo's omitted/rate-0/default medians differ by +5.35/+6.30/+7.19 ns/op and the paired changes are +39.9%/+48.7%/+53.0%. This gap includes the allocator and runtime, not just profiling; the profiling-specific comparisons are the within-toolchain deltas above. LLGo used `GC_disable`; Go used `debug.SetGCPercent(-1)`.

A final same-runner CI comparison ([run](https://github.com/xgo-dev/llgo/actions/runs/32090305786), [full report](https://github.com/xgo-dev/llgo/pull/2027#issuecomment-5152773847)) built both revisions first, then alternated base/head order for every workload and sample. Allocation rows below are differences between 18-process medians; `InterfaceCall` uses the median of seven pairwise signed differences and percentage changes.

| Comparison | Linux/amd64 | Darwin/arm64 |
| --- | ---: | ---: |
| No-consumer allocation run vs PR base | -53.16 ms / -7.5% | -54.22 ms / -8.4% |
| Rate 0 run vs no consumer | +14.37 ms / +2.2% | +5.745 ms / +1.0% |
| Default-rate run vs no consumer | +29.26 ms / +4.5% | +60.92 ms / +10.4% |
| No-consumer build vs PR base | +90.2 ms / +6.1% | +119.4 ms / +6.0% |
| No-consumer file size vs PR base | -4928 B / -0.8% | -320 B / -0.1% |
| `InterfaceCall` vs PR base | -0.028 ns/op / -0.7% | -1.321 ns/op / -23.0% |

The large Darwin `InterfaceCall` improvement was consistent in all seven pairs and comes from shrinking the cold `IfacePtrData` body after the old inlined recorder path was removed. Other unrelated microbenchmark pair ranges crossed zero; scheduler-heavy goroutine measurements remained noisy.

Final no-consumer binaries remove the profiler code. Minimal Linux `cprintf` is nevertheless +256 B on disk because 16 additional pre-DCE funcinfo symbol-index records (16 B each) remain in `.rodata`; `.relro_padding` shrinks by the same 256 B, so `.text`, loaded-section total, and page count are unchanged. Filtering dead funcinfo records is follow-up work.

## Validation

- LLGo `./test/go/memprofile`: concurrent snapshots, reflect closures, tiny allocations, same-stack mixed sizes, and pprof output
- Go GOROOT `test/heapsampling.go`
- targeted `cl`, `ssa`, and `internal/build` tests, including cache separation, library selection, cross-package attribution, and one native TLS resolution
- Darwin/arm64 and Linux/amd64 final-symbol, size, and statement-line checks for the omitted path, plus enabled-path disassembly
- wasip1/wasm build with a direct `runtime.MemProfile` consumer
- normal and full-LTO selection; `c-archive` and `c-shared` retention

Bare-metal target coverage is left to CI. Allocator-specific tuning, sampled-object free accounting, native-equivalent stack capture for wasm/bare-metal, and dead funcinfo filtering remain follow-up work.

### DIFF
--- a/.github/llgo-benchmark.yml
+++ b/.github/llgo-benchmark.yml
@@ -5,14 +5,12 @@ site-path: benchmark/baseline
 include:
   - "^BenchmarkProgram/"
   - "^Benchmark(MergeCompilerFlags|MergeLinkerFlags|LookupPCRandom)$"
-  - "^BenchmarkMemProfile(NoConsumer|Rate0|Default)$"
   - "^Benchmark(RuntimeGetG|Global(Read|Write))$"
   - "^Benchmark(DirectCall|InterfaceCall|Defer|Goroutine)$"
   - "^BenchmarkChannel(Buffered|Handoff)$"
 groups:
   programs: "^Program/"
   compiler: "^(MergeCompilerFlags|MergeLinkerFlags|LookupPCRandom)$"
-  memprofile: "^MemProfile(NoConsumer|Rate0|Default)$"
   core:
     match:
       - "^(RuntimeGetG|Global(Read|Write))$"
@@ -43,20 +41,6 @@ views:
         run-ns:
           title: Run
           format: duration-ns
-  memprofile:
-    title: Memory profiling allocation benchmarks
-    select:
-      groups: "^memprofile$"
-      metrics: "^ns/op$"
-    table:
-      rows: [platform, benchmark]
-      columns: [metric]
-      missing: error
-      empty: error
-      dimensions:
-        benchmark:
-          title: Profile path
-          trim-prefix: BenchmarkMemProfile
   core:
     title: Core language and compiler benchmarks
     select:
--- a/.github/llgo-benchmark.yml
+++ b/.github/llgo-benchmark.yml
@@ -5,12 +5,14 @@ site-path: benchmark/baseline
 include:
   - "^BenchmarkProgram/"
   - "^Benchmark(MergeCompilerFlags|MergeLinkerFlags|LookupPCRandom)$"
+  - "^BenchmarkMemProfile(NoConsumer|Rate0|Default)$"
   - "^Benchmark(RuntimeGetG|Global(Read|Write))$"
   - "^Benchmark(DirectCall|InterfaceCall|Defer|Goroutine)$"
   - "^BenchmarkChannel(Buffered|Handoff)$"
 groups:
   programs: "^Program/"
   compiler: "^(MergeCompilerFlags|MergeLinkerFlags|LookupPCRandom)$"
+  memprofile: "^MemProfile(NoConsumer|Rate0|Default)$"
   core:
     match:
       - "^(RuntimeGetG|Global(Read|Write))$"
@@ -41,6 +43,20 @@ views:
         run-ns:
           title: Run
           format: duration-ns
+  memprofile:
+    title: Memory profiling allocation benchmarks
+    select:
+      groups: "^memprofile$"
+      metrics: "^ns/op$"
+    table:
+      rows: [platform, benchmark]
+      columns: [metric]
+      missing: error
+      empty: error
+      dimensions:
+        benchmark:
+          title: Profile path
+          trim-prefix: BenchmarkMemProfile
   core:
     title: Core language and compiler benchmarks
     select:
--- a/.github/workflows/benchmark-publish.yml
+++ b/.github/workflows/benchmark-publish.yml
@@ -14,7 +14,7 @@ permissions:
 jobs:
   publish:
     if: github.event.workflow_run.conclusion == 'success'
-    uses: xgo-dev/setup-benchmark-go-action/.github/workflows/publish.yml@v1.0.4
+    uses: xgo-dev/setup-benchmark-go-action/.github/workflows/publish.yml@v1
     with:
       run_id: ${{ github.event.workflow_run.id }}
       config_path: .github/llgo-benchmark.yml
--- a/.github/workflows/benchmark-publish.yml
+++ b/.github/workflows/benchmark-publish.yml
@@ -14,7 +14,7 @@ permissions:
 jobs:
   publish:
     if: github.event.workflow_run.conclusion == 'success'
-    uses: xgo-dev/setup-benchmark-go-action/.github/workflows/publish.yml@v1
+    uses: xgo-dev/setup-benchmark-go-action/.github/workflows/publish.yml@v1.0.5
     with:
       run_id: ${{ github.event.workflow_run.id }}
       config_path: .github/llgo-benchmark.yml
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -78,12 +78,14 @@ jobs:
             "$GITHUB_WORKSPACE/.benchmark/results"
 
       - name: Record benchmark result
-        uses: xgo-dev/setup-benchmark-go-action@v1.0.4
+        uses: xgo-dev/setup-benchmark-go-action@v1
         with:
           config: .github/llgo-benchmark.yml
           benchmark-file: .benchmark/results/benchmark.txt
           baseline-benchmark-file: >-
             ${{ github.event_name == 'pull_request' &&
                 '.benchmark/base-results/benchmark.txt' || '' }}
+          sample-pairing: >-
+            ${{ github.event_name == 'pull_request' && 'index' || '' }}
           platform-id: ${{ matrix.id }}
           platform-label: ${{ matrix.display }}
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -38,7 +38,16 @@ jobs:
         with:
           repository: ${{ github.event.pull_request.base.repo.full_name }}
           ref: ${{ github.event.pull_request.base.sha }}
-          path: .benchmark/source
+          path: .benchmark/base-source
+          persist-credentials: false
+
+      - name: Check out pull request head benchmark source
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v7
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          path: .benchmark/head-source
           persist-credentials: false
 
       - name: Install dependencies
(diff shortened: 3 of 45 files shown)


cpunion:
Rebased the independent memprofile root onto current `main` (`551626a1d`) and force-pushed the branch.

Conflict handling kept current-main panic/recover snapshot hooks in `unwind_llgo.go`, then layered the memprofile initialization/capture hooks beside them. The final xfail change removes only native `heapsampling.go` entries; there are no CI timing/config changes and no wasm behavior work.

Fresh validation:
- compiler detection/instrumentation tests pass; `packageReadsMemProfile` coverage 95.7%
- runtime modules build
- LLGo memprofile and caller/panic/introspection acceptance tests pass
- upstream `heapsampling.go` passes without xfail with Go 1.24.11, 1.25.0, and 1.26.5 on darwin/arm64

cpunion:
Benchmark publishing is now pinned to `setup-benchmark-go-action@v1.0.5` in both places: the PR-side recorder and the default-branch `workflow_run` publisher.

Validation used real paired LLGo artifacts. The v1.0.5 renderer reports the measured base SHA and signed unit/percentage deltas (for example, `+256 B / +1.3%` and `0 ns/op / +0.0%`); index-paired samples are summarized with pairwise medians.

The live bot comment on this PR is still produced by the publisher currently on `main` (v1.0.4), so its presentation remains percentage-only until this workflow change is merged. After merge, the default-branch publisher is pinned to v1.0.5 as well.

PR #2033, opened 2026-07-06
ssa,runtime: tighten recover to direct deferred calls

cpunion:
## Problem

Go permits `recover()` only when called directly by the deferred function currently handling the panic. An indirect helper, nested closure, or recursive sub-call must receive nil. The prior runtime could not reliably distinguish those calls.

## Implementation

- associate each panic node with its owning `Defer` frame and check that ownership in `Recover`;
- extend the current `g` with recover-frame ownership;
- preserve closure and method-value wrapper transparency with recover-frame aliases;
- keep the existing unrecovered `PanicTraceback` hook and report the panic node argument;
- mark may-recover LLVM functions with a module-local string attribute instead of a process-global map.

LLVM value handles can be reused after module disposal, so a global recover-classification cache could leak stale state into a later package. This PR adds no pthread key or second per-goroutine store.

The independent diff is `+624/-151` across 16 files.

## Conformance

Removed xfails for `fixedbugs/issue4066.go`, `fixedbugs/issue73916.go`, and `fixedbugs/issue73916b.go`.

## Validation

- macOS arm64, Go 1.26.5: affected SSA/compiler coverage tests and real LLGo fixedbug recover tests pass.
- Full CI-style Go test run passes, including `test/go` and the standard-library coverage packages.
- The generated main coverage profile reports 75.18% statement coverage; focused recover branches are covered separately.
- Runtime module compilation and host Go semantics tests pass.

This supersedes the remaining recover-ownership portion of #1918.

### DIFF
--- a/cl/_testgo/cgodefer/cgodefer.go
+++ b/cl/_testgo/cgodefer/cgodefer.go
@@ -6,6 +6,11 @@ package main
 */
 import "C"
 
+func main() {
+	p := C.malloc(1024)
+	defer C.free(p)
+}
+
 // CHECK-LABEL: define [0 x i8] @main._Cfunc_free(ptr %0){{.*}} {
 // CHECK-NEXT: _llgo_0:
 // CHECK-NEXT:   %1 = call ptr @"{{.*}}/runtime/internal/runtime.AllocZ"(i64 8)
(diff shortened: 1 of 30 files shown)


cpunion:
Rebased onto current `main` (`551626a1d`). The conflicts were resolved by preserving the current per-G panic traceback state and adding the recover-frame ownership field/logic alongside it; the xfail removals were reapplied against the current YAML rather than carrying the old file forward.

Revalidated:
- full `go test ./ssa -count=1`
- affected `cgodefer` and `recoverthenpanic` compiler/runtime tests
- LLGo `test/go` recover/panic tests
- Go 1.24.11, 1.25.0, and 1.26.5 `fixedbugs/issue4066.go` with empty xfail
- Go 1.26.5 `issue73916.go` and `issue73916b.go` with empty xfail

cpunion:
Superseded by #2034. The direct-defer, activation-identity, and wrapper-transparency changes are preserved there as separate layered commits. The Dev LTO/GlobalDCE regression has also been fixed and covered by a focused test.

PR #2044, opened 2026-07-08
ci: one heavy test suite per runner VM

cpunion:
Structural fix for the post-backbone runner deaths (ubuntu exit-143 "shutdown signal" ~8 minutes into the heavy-package phase, reproduced on main and every stage-5 branch) and the mac go-list `WaitDelay` flakes: the covered run stacks four heavy llgo-building test packages (cl, ssa, internal/cabi, test/go) on one VM, and warm shared caches let their compile phases fully overlap.

## Design (asymmetric, mindful of scarce mac runners)

- **ubuntu** (plentiful runners, where the VMs die): one suite per VM — `cl` / `ssa+cabi` / `test tree` / `rest`. Peak cannot stack by construction.
- **macos** (scarce runners; symptom was slowness, never VM death): two balanced groups of at most two heavy packages each — `cl+cabi` and `ssa+testgo+rest`. That restores the pre-backbone peak scale, which mac handled, at the cost of only +1 mac job.

Rejected alternatives: raising budgets (hides it), `go test -p 2` (lowers probability only).

Wall time drops to the slowest group; coverage uploads from every job and codecov merges per commit. The embedded-emulator step rides with the cl suites, the std-symbol check with the rest suites.

**Branch protection note:** required check names change from `test (os, 19)` to `test (os, 19, <suite>)`.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,14 +18,22 @@ concurrency:
 
 jobs:
   test:
+    name: test (${{matrix.os}}, ${{matrix.llvm}}, ${{matrix.suite}})
     continue-on-error: true
     timeout-minutes: 60
     strategy:
+      fail-fast: false
       matrix:
         os:
           - macos-latest
           - ubuntu-latest
         llvm: [19]
+        # One heavy test package per runner VM: their compile phases no
+        # longer stack on a single 16GB machine (concurrent clang/lld
+        # bursts were killing ubuntu runners and starving mac runners
+        # into go-list WaitDelay expirations), and wall time drops to the
+        # slowest single suite.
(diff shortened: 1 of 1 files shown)


cpunion:
Closing: the runner-death root cause turned out to be per-process memory accumulation in the heavy test binaries (co-resident anon peak ~6.2GB), fixed at the source in #2048 — whose CI runs the unsplit combined go-test jobs green on both platforms (47/47). With the fix, suite co-resident anon is ~2.9GB, below pre-#2024 levels, so the structural split is no longer needed and we avoid the branch-protection check-name churn. The wall-clock benefit of splitting (mac coverage 45m→14m) can be re-proposed later as a pure speed improvement, decoupled from stability.

PR #2056, opened 2026-07-10
ssa: avoid unused unexported method names

luoliwoshang:
Avoid materializing an unexported method short name before replacing it with its ABI-qualified name.

Refreshes ordinary source-embedded FileCheck directives with chore/litgen and retains hand-written captures plus CHECK-NOT assertions for ABI metadata facts that litgen cannot infer.

Tested:
- go test -v -covermode=atomic ./cl -run "^TestRunAndTestFromTestgo/(abimethod|tpinst)$" -count=1


luoliwoshang:
@xushiwei

Issue #1608, opened 2026-02-04
llgo: Slice return value corrupted when struct field modified after read


luoliwoshang:
## Problem Description

When a function returns a single slice value, and the slice is read from a struct field, then that field is modified before the return, llgo incorrectly returns the modified value instead of the original value.

**This affects all control flow statements with initialization** (if, switch, type switch), as they internally use this pattern in gogen's `clearBlockStmt()` function.

## Minimal Reproduction

```go
package main

type T struct {
	data []string
}

// ❌ Single return value - BROKEN in llgo
func a() []string {
	var t = T{data: []string{"a", "b"}}
	a := t.data                      // Read: ["a", "b"]
	t.data = []string{"a", "b", "c"} // Modify field
	return a                          // Expected: ["a", "b"], Actual: ["a", "b", "c"]
}

// ✅ Multiple return values - WORKS in llgo
func b() ([]string, bool) {
	var t = T{data: []string{"a", "b"}}
	a := t.data                      // Read: ["a", "b"]
	t.data = []string{"a", "b", "c"} // Modify field
	return a, true                    // Correctly returns ["a", "b"]
}

func main() {
	println(a()) // llgo: [3/3] ❌ Wrong length!
	println(b()) // llgo: [2/2] ✅ Correct
}
```

## Expected Behavior

**Standard Go output:**
```
[2/2]0xc0000...  # func a() returns length 2
[2/2]0xc0000...  # func b() returns length 2
```

## Actual Behavior (llgo)

**llgo output:**
```
[3/3]0xc0000...  # func a() returns length 3 ❌
[2/2]0xc0000...  # func b() returns length 2 ✅
```

### LLVM IR shows the issue

The generated IR for single-return-value function appears correct, but at runtime the return value gets overwritten:

```llvm
define %Slice @"...a"() {
  %9 = load %Slice, ptr %8, align 8      ; Load original [a, b]
  %16 = insertvalue %Slice ..., i64 3, 2 ; Create new [a, b, c]
  store %Slice %16, ptr %17, align 8     ; Store new value
  ret %Slice %9   ; Return %9, but it contains [a, b, c]!
}
```

### Why multi-value return works

The multi-value return uses `insertvalue` to construct a struct, which **creates a copy** of the slice:

```llvm
define { %Slice, i1 } @"...b"() {
  %9 = load %Slice, ptr %8, align 8
  store %Slice %16, ptr %17, align 8
[...]

PR #2103, opened 2026-07-16
cabi: bound large aggregate return lowering

cpunion:
## Summary

- add a minimal ARM64 ABI fast path before flattening aggregates, while preserving HFA register returns
- keep large indirect returns in memory in the existing sret conversion paths instead of materializing giant LLVM SSA loads/stores
- remove only exact, non-volatile large self-copies encountered while transforming sret function bodies
- retire the Go 1.26 `fixedbugs/issue15141.go` compile xfail

## Reproducer

The Go 1.26 `fixedbugs/issue15141.go` case returns a 268,435,455-byte array in several ordinary source forms:

```go
func f(i, y int) (a [0xFFFFFFF]byte) {
    a[i] = byte(y)
    return
}

func g(i, y int) [0xFFFFFFF]byte {
    var a [0xFFFFFFF]byte
    a[i] = byte(y)
    return a
}

func h(i, y int) (a [0xFFFFFFF]byte) {
    a[i] = byte(y)
    return a
}
```

All three forms should compile with an indirect ABI return without constructing the array as an element-by-element LLVM value.

## Root cause

The ARM64 classifier previously flattened every aggregate before checking whether its size required indirect passing:

```go
types := elementTypes(p.td, typ)
n := len(types)
// HFA and other small-aggregate checks...
if info.Size > 16 {
    info.Kind = AttrPointer
    info.Type1 = llvm.PointerType(typ, 0)
}
```

For `[0xFFFFFFF]byte`, `elementTypes` therefore attempted to build a slice containing 268,435,455 LLVM types before the existing `Size > 16` branch could select `AttrPointer`. The ABI result was already correct, but reaching it exhausted the bounded compile-time memory budget.

## Design

The `arch.go` change is limited to a five-line ARM64 fast path: aggregates larger than 16 bytes with more than four scalar elements return `AttrPointer` before flattening. The original classifier remains unchanged for all other values, including ARM64 HFAs with up to four elements. Other ABI classifiers and the existing `elementTypes` implementation are unchanged.

The large-copy handling is also local to C ABI indirect returns; there is no module-wide load/store pass. On the callee side, a single-use return load is copied to sret at the original load position, preserving the value before a later source mutation as required by #1608. On the caller side, store-only uses are copied directly from the stable sret temporary. Exact same-address load/store pairs in an sret body are removed as memory no-ops. Volatile or otherwise unsupported uses retain the existing behavior.

`maxDirectAggregateCopySize` (1 MiB) is an LLGo implementation heuristic, not a Go language, ABI, or compiler requirement. It is deliberately far below the 256 MiB reproducer while remaining above ordinary ABI aggregates, so this resource safeguard fixes the pathological LLVM scalarization without broadly changing existing small-value lowering. The Go compiler instead uses its structural `CanSSA` policy and turns non-SSA aggregate copies into memory `Move` operations; adopting that broader policy in LLGo would affect many smaller arrays and structs and is outside this focused fix.

## Validation

- Darwin/arm64 and Linux/amd64 with Go 1.26.5: `fixedbugs/issue15141.go` passes under a 1536 MiB RSS hard limit without reaching the 256 MiB warning threshold
- complete `go test ./internal/cabi` passes on Darwin/arm64 and Linux/amd64
- Darwin package statement coverage: 96.8%; the new aggregate classification helpers and `transformFuncBody` are 100% covered
- regression IR verifies copy-before-mutation semantics, direct use of the sret temporary, exact self-copy removal, HFA preservation, and unchanged volatile operations
- `go vet ./internal/cabi` and `git diff --check` pass

### DIFF
--- a/internal/cabi/arch.go
+++ b/internal/cabi/arch.go
@@ -192,19 +192,19 @@ func (p *TypeInfoArm64) GetTypeInfo(ctx llvm.Context, ftyp llvm.Type, typ llvm.T
 	info.Align = p.Alignof(typ)
 	switch kind {
 	case llvm.StructTypeKind, llvm.ArrayTypeKind:
-		types := elementTypes(p.td, typ)
-		n := len(types)
+		n := elementTypesCount(typ)
 		if bret && n == 1 {
 			return info
 		}
-		if n == 2 {
-			// skip (i64/ptr,i64/ptr)
-			if (types[0].TypeKind() == llvm.PointerTypeKind || types[0] == ctx.Int64Type()) &&
-				(types[1].TypeKind() == llvm.PointerTypeKind || types[1] == ctx.Int64Type()) {
-				return info
-			}
-		}
 		if n <= 4 {
+			types := elementTypes(p.td, typ)
+			if n == 2 {
+				// skip (i64/ptr,i64/ptr)
+				if (types[0].TypeKind() == llvm.PointerTypeKind || types[0] == ctx.Int64Type()) &&
+					(types[1].TypeKind() == llvm.PointerTypeKind || types[1] == ctx.Int64Type()) {
+					return info
+				}
+			}
 			if checkTypes(types, ctx.FloatType()) || checkTypes(types, ctx.DoubleType()) {
 				return info
 			}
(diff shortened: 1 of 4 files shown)


cpunion:
Replacement implementation: #2123. The CABI-local lowering here bounds compilation in ABI mode 2, but the full `issue15141` executable still places 256 MiB locals on the native stack and fails at runtime; ABI mode 0 also bypasses this path. #2123 moves the fix before CABI, heapifies oversized locals, and adds ABI 0/1/2 execution coverage. I will close this PR as superseded after the replacement CI finishes.

cpunion:
Superseded by #2123, which implements the fix before CABI so ABI modes 0/1/2 are covered and has now been merged into main.

PR #2106, opened 2026-07-16
ssa: implement -B bounds-check disabling

cpunion:
## Summary

- accept the gc-compatible `llgo tool compile -B` flag and propagate it through build configuration into SSA lowering
- omit index, string/slice, two/three-index slice, and slice-to-array conversion bounds checks when enabled
- retain integer-width conversions, required nil checks for `*array`, and mandatory `unsafe.String` / `unsafe.Slice` validity checks
- include the setting in build fingerprints and keep the default checked path unchanged
- retire the Go 1.26 `fixedbugs/issue48092.go` compile xfail

## Scope

This change implements the compiler-action `-B` used by GOROOT tests and an explicit build `Config` setting. It does not reinterpret top-level `llgo build/run/test -gcflags=-B`: Go gcflags are package-pattern scoped with last-match replacement semantics, which require a separate per-package design.

The compatibility command accepts `-o`, but its compile-only path validates lowering without emitting an object file. This PR preserves that behavior; output-file compatibility is outside its scope.

## Validation

Rebased on current `xgo-dev/llgo main` (`73fa30fbf`):

- macOS arm64, Go 1.26.5:
  - complete `cmd/internal/compile`, `ssa`, and `internal/build` suites pass
  - coverage: 97.4%, 87.9%, and 73.4%; combined 80.8%
  - GOROOT `fixedbugs/issue48092.go` passes with its xfail removed
- Linux amd64, LLVM 19, Go 1.26.5:
  - complete `cmd/internal/compile`, `ssa`, and `internal/build` suites pass
  - coverage: 97.4%, 88.0%, and 72.6%
  - GOROOT `fixedbugs/issue48092.go` passes
- key lowering coverage: `checkIndex` 100%, `Slice` 98.6%, `SliceToArrayPointer` 100%, and `DisableBoundsChecks` 100%
- checked and unchecked programs retain identical legal results; IR tests prove bounds helpers disappear under `-B`
- negative tests retain required `*array` nil panics and `unsafe.String` / `unsafe.Slice` validity panics
- `git diff --check` passes

The previous Ubuntu LLGo failure was unrelated to `-B`: it exercised the former main behavior that rejected DWARF omission for c-shared/c-archive. That support is present in current main and the rebased CI reruns it.

### DIFF
--- a/cmd/internal/compile/compile.go
+++ b/cmd/internal/compile/compile.go
@@ -162,6 +162,7 @@ func runCmd(_ *base.Command, args []string) {
 	conf.GoVersion = opts.lang
 	conf.NoErrorColumn = opts.noColumns.value != 0
 	conf.AllowNoBody = !opts.complete
+	conf.DisableBoundsChecks = opts.noBounds.value != 0
 	var loaderCompilerFlags []string
 	if opts.allErrors.value != 0 {
 		loaderCompilerFlags = append(loaderCompilerFlags, "-e")
@@ -198,7 +199,6 @@ func (opts *options) unsupported() []string {
 			out = append(out, name)
 		}
 	}
-	appendFlag(opts.noBounds.value != 0, "-B")
 	appendFlag(opts.dynlink, "-dynlink")
 	appendFlag(opts.showOpt.value != 0, "-m")
 	appendFlag(opts.live, "-live")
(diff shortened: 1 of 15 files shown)


xushiwei:
@gemini-code-assist /review

PR #2146, opened 2026-07-22
build: preserve DebugRefs during SSA order repairs

### LINKED ISSUE #2122: DWARF: preserve crypto/x509 OID.Equal semantics
## Summary

Enabling LLGo's current DWARF path changes executable behavior in `crypto/x509`: `TestOID` reports that two equal OIDs compare unequal. Omitting DWARF makes the same test pass.

This is a correctness bug, not merely incomplete debug metadata, and is one reason broad builds cannot safely enable the current DWARF path by default yet.

## Reproduction

Confirmed locally on native Linux/arm64 with Go 1.26.5 and the current compiler from https://github.com/xgo-dev/llgo/pull/2120:

```sh
llgo test -ldflags=-w=false -run '^TestOID$' github.com/goplus/llgo/test/std/crypto/x509
```

Result:

```text
--- FAIL: TestOID (0.00s)
    x509_test.go:175: OID.Equal returned false for equal OIDs
FAIL
```

Control:

```sh
llgo test -ldflags=-w -run '^TestOID$' github.com/goplus/llgo/test/std/crypto/x509
```

Result: `PASS`.

The same failure appears in the Go 1.24 Linux shard of https://github.com/xgo-dev/llgo/actions/runs/29549966202, so it is not specific to the local Go 1.26 toolchain.

## Analysis

`-w=false` currently selects more than section emission. `internal/build` enables `ssa.GlobalDebug`, disables the LLVM pass pipeline, and disables C ABI optimization before asking the linker to keep DWARF. The differential test therefore isolates the bug to the DWARF-enabled build path, but does not yet prove that the metadata nodes themselves corrupt execution.

Likely investigation split:

1. independently toggle `ssa.GlobalDebug`, `passOpt`, and `cabiOptimize` for this test;
2. compare IR for `encoding/asn1.OID.Equal` and its slice/loop operations;
3. determine whether the failure is debug-mode SSA lowering, unoptimized C ABI lowering, or a pass dependency that currently masks invalid IR;
4. add a smaller OID/slice equality reproducer before changing the DWARF implementation.

The ignored `net/url.(*URL).setPath` linkname warning appears in both failing and passing controls and is therefore not the discriminator.

## Expected behavior

Debug information must not change Go program semantics. `OID.Equal` must return true for equal values with both `-w=false` and `-w`.

cpunion:
## Problem

LLGo repairs Go SSA evaluation order in two places: return values whose storage is mutated by a later call, and assignment targets for single-case select receives. With `ssa.GlobalDebug`, `ssa.DebugRef` metadata also uses the affected values. Treating those metadata uses as executable dependencies blocks both otherwise-safe repairs.

That makes enabling DWARF change program behavior:

- `crypto/x509.ParseOID` can return a stale OID because its result load stays before the mutating call.
- A single-case select can evaluate its assignment target before the receive; the acceptance test observes `[100 1]` instead of `[1 100]`.

## Changes

- move a repaired return load and its `ssa.DebugRef` users as one stable group after the mutating call
- move `ssa.DebugRef` users together with single-case select assignment-target dependencies after the receive
- continue rejecting either transformation when an executable instruction would observe a moved value too early

The change is confined to the existing SSA order-repair module. It does not alter the loader, DWARF encoding, C ABI lowering, or LLVM optimization pipeline.

Fixes #2122.

## Verification

- synthetic pointer, map, and two-value select regressions with default SSA and `ssa.GlobalDebug`
- plain receive and multi-case select non-regressions
- return-order regressions with default SSA and `ssa.GlobalDebug`
- direct `crypto/x509.ParseOID` SSA regression
- exact `TestSingleCaseSelectReceiveAssignOrder` acceptance with `-w=false` on macOS and Ubuntu arm64
- `crypto/x509 TestOID` with `-w=false` on macOS and Ubuntu arm64
- full `internal/build` package on macOS
- 100% statement coverage for the new select DebugRef grouping helper

### DIFF
--- a/internal/build/ssa_order_fix.go
+++ b/internal/build/ssa_order_fix.go
@@ -302,11 +302,15 @@ func fixSSAOrderBlock(b *ssa.BasicBlock) {
 			continue
 		}
 
-		// If the loaded value is used by any instruction between its current
-		// position and the return (excluding return itself), moving it may place
-		// its definition after one of those uses and break SSA form.
+		// DebugRefs are metadata-only and move with the value they describe. Any
+		// executable use before Return still makes reordering unsafe.
+		moving := map[ssa.Instruction]struct{}{u: {}}
 		usedBeforeReturn := false
 		for i := loadIdx + 1; i < retIdx; i++ {
+			if ref, ok := b.Instrs[i].(*ssa.DebugRef); ok && instrUsesValue(ref, u) {
+				moving[ref] = struct{}{}
+				continue
+			}
 			if instrUsesValue(b.Instrs[i], u) {
 				usedBeforeReturn = true
 				break
@@ -316,9 +320,7 @@ func fixSSAOrderBlock(b *ssa.BasicBlock) {
 			continue
 		}
 
-		// Move the load right after the last call (but before Return).
-		b.Instrs = moveInstr(b.Instrs, loadIdx, lastCallIdx+1)
-		// Adjust retIdx for subsequent moves in this block.
+		b.Instrs = moveInstrsAfter(b.Instrs, moving, b.Instrs[lastCallIdx])
 		retIdx = indexOfInstr(b.Instrs, ret)
 	}
 }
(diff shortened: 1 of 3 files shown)


cpunion:
Consolidated into #2143, which now contains the SSA DebugRef order-repair changes and their regression coverage. Closing this redundant PR; no code is being merged to main by this action.

PR #2159, opened 2026-07-22
lto: analyze aggregate MethodByName string arguments

cpunion:
Unblocks https://github.com/xgo-dev/llgo/pull/2143.

## Problem

The MethodByName LTO pass assumed that every marked Go string argument had been lowered into separate `(ptr, len)` parameters. LLGo's arm64 LP64 C ABI path can instead keep the 16-byte Go string as a `{ptr, len}` aggregate. In that form the pass could not recover direct, concatenated, sliced, or forwarded constant names, so it left the generic reflect marker in place and prevented the intended named-method pruning.

## Changes

- consume one logical string argument in either aggregate or split ABI form
- reuse that decoding for direct MethodByName calls, `runtime.StringCat`, `runtime.StringSlice2`, and forwarded function parameters
- preserve the generic marker whenever a name cannot be resolved
- add Go fixtures covering direct, concat, slice, forwarding, linked execution, actual unused-method pruning, and unknown-name fallback
- derive the aggregate-ABI assertions from generated Go modules rather than hand-written LLVM IR

The focused aggregate test generates frontend IR with automatic C ABI lowering disabled, applies LLGo's arm64 C ABI transformer, and runs the LTO pass. One module checks the four known-name markers without a generic value marker. A separate module checks conservative unknown-name fallback because a generic marker intentionally retains all matching methods and would invalidate an unused-method symbol assertion. The normal plugin runtime and symbol suites exercise the linked split-ABI path and verify that `Known.Drop` is absent from the final binary.

This PR changes analysis only; it does not change either ABI.

## Validation

- negative check: the Go-generated aggregate test fails with the `main` plugin at the missing `Concat` marker and passes with this PR
- macOS arm64, LLVM 19: clean plugin build and both generated aggregate modules pass
- Ubuntu arm64 container, capped at 2 CPUs / 15 GiB: both generated aggregate modules pass
- Ubuntu arm64: linked runtime fixture passes and the final symbol table excludes `Known.Drop`
- Ubuntu arm64: complete existing LTO-plugin runtime and symbol suites; the new fixture passes in the runtime suite
- Ubuntu arm64: CI-equivalent plugin coverage command, 52.7% aggregate coverage for the instrumented Go packages
- prior CI: `Dev LTO GlobalDCE`, both coverage platforms, and Codecov patch coverage pass

### DIFF
--- a/cl/_testlto/globaldce_reflect_method_by_name_ltoplugin_string_abi/in.go
+++ b/cl/_testlto/globaldce_reflect_method_by_name_ltoplugin_string_abi/in.go
@@ -1,0 +1,53 @@
+// LITTEST
+package main
+
+import (
+	"reflect"
+)
+
+// SYMBOL-NOT: globaldce_reflect_method_by_name_ltoplugin_string_abi{{.*}}Known{{.*}}Drop
+// SYMBOL-DAG: globaldce_reflect_method_by_name_ltoplugin_string_abi{{.*}}Known{{.*}}Direct
+// SYMBOL-DAG: globaldce_reflect_method_by_name_ltoplugin_string_abi{{.*}}Known{{.*}}Concat
+// SYMBOL-DAG: globaldce_reflect_method_by_name_ltoplugin_string_abi{{.*}}Known{{.*}}Slice
+// SYMBOL-DAG: globaldce_reflect_method_by_name_ltoplugin_string_abi{{.*}}Known{{.*}}Forward
+// SYMBOL-NOT: globaldce_reflect_method_by_name_ltoplugin_string_abi{{.*}}Known{{.*}}Drop
+
+type Known struct{}
+
+//go:noinline
+func (Known) Direct() string { return "direct" }
+
+//go:noinline
+func (Known) Concat() string { return "concat" }
+
+//go:noinline
+func (Known) Slice() string { return "slice" }
+
+//go:noinline
+func (Known) Forward() string { return "forward" }
+
+//go:noinline
+func (Known) Drop() string { panic("unreachable") }
+
+func callForward(name string) string {
+	out := reflect.ValueOf(Known{}).MethodByName(name).Call(nil)
+	return out[0].String()
+}
+
+func callConcat(prefix, suffix string) string {
+	out := reflect.ValueOf(Known{}).MethodByName(prefix + suffix).Call(nil)
+	return out[0].String()
+}
+
+func callSlice(source string) string {
+	out := reflect.ValueOf(Known{}).MethodByName(source[2:7]).Call(nil)
+	return out[0].String()
+}
+
+func main() {
+	v := reflect.ValueOf(Known{})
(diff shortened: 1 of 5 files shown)


cl/_testlto/globaldce_reflect_method_by_name_ltoplugin_string_abi/in.go
```diff
@@ -0,0 +1,62 @@
...
+func (Known) Concat() string { return "concat" }
+
+//go:noinline
+func (Known) Slice() string { return "slice" }
+
+//go:noinline
+func (Known) Forward() string { return "forward" }
+
+//go:noinline
+func (Known) Drop() string { panic("unreachable") }
```
zhouguangyuan0718, line 30: 这个用例要检查下确保Drop确实被删除了

ltoplugin/LLGOReflectMethodByNamePass.cpp
```diff
@@ -173,20 +173,40 @@ bool isRuntimeStringSlice2(CallBase *CB) {
...
 
   SmallVector<std::string, 4> Prefixes;
   SmallVector<std::string, 4> Suffixes;
-  if (!collectStringSet(CB->getArgOperand(0), CB->getArgOperand(1), DL,
-                        Prefixes, Depth + 1) ||
-      !collectStringSet(CB->getArgOperand(2), CB->getArgOperand(3), DL,
-                        Suffixes, Depth + 1))
+  unsigned ArgIndex = 0;
+  if (!collectStringSetFromCallArgument(CB, ArgIndex, DL, Prefixes, Depth) ||
+      !collectStringSetFromCallArgument(CB, ArgIndex, DL, Suffixes, Depth) ||
```
zhouguangyuan0718: 这里两次调用传的ArgIndex一样 可能不太对

cpunion: 这里是有意复用同一个按引用推进的参数游标：第一次消费 aggregate 时推进 1、消费 split `(ptr, len)` 时推进 2，第二次从后续参数继续。为避免误读，已将 helper/变量重命名为 `consumeStringCallArgument` / `NextArgIndex`；Concat 的聚合 ABI 测试会同时验证两次消费组合出 `Concat`。macOS 和 Ubuntu 的聚合测试均已通过。

cl/_testlto/globaldce_reflect_method_by_name_ltoplugin_string_abi/in.go
cpunion: 已补充最终链接符号检查，明确断言 `Known.Drop` 不存在。加断言后还发现原 fixture 中的 unknown-name generic marker 会按设计保守保留所有匹配方法，因此把 unknown fallback 拆到了独立 module；Ubuntu 的链接符号测试现已确认 `Known.Drop` 从最终 ELF 中删除。

zhouguangyuan0718 (approved):

PR #2174, opened 2026-07-24
build: prepare bounded parallel compilation

zhouguangyuan0718:
## Summary
- map Go build `-p` to typed LLGo parallelism and use it for the SSA build worker pool
- split package frontend/backend preparation while keeping LLVM backend execution serial
- make cache dependency collection include alternate-package imports and initialize package build state explicitly

## Validation
- `go test ./internal/build ./internal/goflags ./cl -count=1`

This is the first PR in the consolidated package-parallel build stack.

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -183,6 +183,9 @@ type Config struct {
 	// DisableBoundsChecks disables index, slice, and slice-to-array conversion
 	// bounds checks while retaining required integer conversions and nil checks.
 	DisableBoundsChecks bool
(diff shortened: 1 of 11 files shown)


zhouguangyuan0718:
Superseded by the rebuilt stack starting at #2175. The old PR mixed SSA worker preparation with an obsolete dependency-level scheduling boundary; the new stack keeps LLVM serial through #2180 and enables package-level LLVM workers atomically in #2182.

PR #2176, opened 2026-07-24
build: split package execution stages

zhouguangyuan0718:
## Summary
- split the serial package pipeline into preflight, execution, and finalization stages
- keep cache lookup outside LLVM module work and keep LLVM backend execution serial
- centralize archive publication and cache metadata finalization

## Validation
- `GOCACHE=/private/tmp/llgo-pr2-gocache go test ./internal/build ./internal/goflags ./cmd/internal/flags -count=1`

## Dependency
This PR includes and depends on #2174 and #2175. Merge/rebase it after those PRs; it will then contain only this stage split.


zhouguangyuan0718:
This PR has been absorbed into #2175 during the parallel-build stack consolidation. The retained PR now contains the explicit build units and preflight/execute/finalize stage split.

PR #2177, opened 2026-07-24
build: plan package dependency levels

zhouguangyuan0718:
## Summary
- build an immutable package dependency plan from effective dependencies
- derive stable ready levels for a future bounded worker scheduler
- validate duplicate package IDs and dependency cycles without changing the current serial execution order

## Validation
- `GOCACHE=/private/tmp/llgo-pr3-gocache go test ./internal/build ./internal/goflags ./cmd/internal/flags -count=1`

## Dependency
This PR includes and depends on #2174, #2175, and #2176. Merge/rebase it after those PRs; it will then contain only this dependency-plan stage.


zhouguangyuan0718:
Closing as obsolete after the architecture review: LLVM package backends no longer wait on dependency levels. The dependency-DAG/ready-level scheduler has been removed from the consolidated implementation.

PR #2180, opened 2026-07-25
build: isolate package backend sessions

zhouguangyuan0718:
## Summary

- centralize backend Program construction in a build-local template
- create independent backend sessions with their own `llssa.Program`, LLVM Context, TargetMachine, and C ABI transformer
- replay runtime, Python, syntax/directive, and patched-package state into each new Program
- freeze prepared locality declaration/layout state from the main Program after preflight and share it directly with backend Programs, without per-session map copies
- keep package-syntax parse markers Program-local, so workers record their own parsing without mutating the shared locality state
- permit only idempotent re-declaration of an already prepared name/owner pair after freeze; reject genuinely new locality mutations
- precompute caller tracking before it becomes read-only worker input
- preserve the current serial package backend execution in this layer; sessions are infrastructure only until #2182

This layer intentionally has no `PackageSummary` and does not enable package workers. It only makes a complete package-local backend session reproducible without process-global state or shared LLVM-owned objects. The shared locality state contains no LLVM objects and becomes immutable before backend execution.

## Dependency and review boundary

Based directly on current `main`; it no longer depends on #2179.

- base: `ca0d4199c`
- head: `bfdcaae86`
- 7 files, +514 / -34
- 3 commits

Please review `ca0d4199c..bfdcaae86`.

## Tests

- `go test ./internal/build ./cl ./ssa`
- focused backend-session, concurrent Invocation, package test, caller-tracking, and frozen-locality `-race` tests
- repeated frozen shared-locality race test
- `git diff --check`

### DIFF
--- a/cl/caller_tracking_precompute_test.go
+++ b/cl/caller_tracking_precompute_test.go
@@ -1,0 +1,79 @@
+//go:build !llgo
+
+/*
+ * Copyright (c) 2026 The XGo Authors (xgo.dev). All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


zhouguangyuan0718:
Superseded by #2282 followed by #2182. The replacement prepares Program-side syntax/locality/caller metadata once, shares it directly read-only, and creates fresh LLVM-owned state per backend Program. It removes this PRs backend template, freeze/replay, and post-freeze locality mutation machinery. The restacked parallel implementation has passed the affected full suites, testdrop, p=1/p=8 cache tests, concurrent-session tests, and targeted race validation.

PR #2200, opened 2026-07-28
ci: add lightweight Linux and macOS baseline benchmarks

cpunion:
## Summary

- add a bounded per-commit benchmark suite on Linux and macOS
- measure three existing workloads: `cprintf`, `println`, and `fmtprintf`
- record warm build time, process time, file/text/data/bss size, compiler helpers, and core LLGo operations
- export all measurements in standard Go benchmark format
- publish comments, branch/PR/main history, and GitHub Pages through [`setup-benchmark-go-action@v1`](https://github.com/xgo-dev/setup-benchmark-go-action/releases/tag/v1.0.3)

## Benchmark scope

The core stream covers compiler flag merging, PCLN lookup, global access, direct/interface calls, defer, `getg`, buffered/handoff channels, and goroutine creation. TLS/GLS are intentionally not included because those directives are not in `main` yet.

The goroutine benchmark launches a bounded batch with a `sync.WaitGroup`, so it measures creation/completion without adding one serialized channel rendezvous per operation. CI fixes it at 100 iterations because the current native backend creates one pthread per goroutine.

## Publishing and security

PR jobs receive no benchmark data token. They collect results and upload schema-validated artifacts. A trusted default-branch `workflow_run`:

1. downloads all platform artifacts without executing the measured revision;
2. binds repository, SHA, ref, URLs, and timestamp to the workflow event;
3. requires fork PR configuration to exactly match `.github/llgo-benchmark.yml` on `main`;
4. revalidates layouts, units, samples, medians, platforms, and limits;
5. verifies that the PR still points to the measured SHA before writing data or updating its single bot comment.

History and the generated site are committed atomically to `xgo-dev/llgo-benchmark-data`'s `pages` branch. A PR is publishable before any matching `main` measurement exists; every delta is then shown as `new`.

## Cost controls

- `GOMAXPROCS=2` and `go build -p=1`
- three warm build samples and seven process samples per workload
- 250 ms core benchmarks
- goroutine creation fixed at 100 operations
- one Linux and one macOS job; runtime measurements are not cross-compiled

## Validation

- `go test -p=1 -cover ./benchmark/baseline`: 92.1%
- both workflows pass `actionlint` 1.7.7
- local macOS collector, all selected LLGo benchmarks, fixed goroutine benchmark, export, record, and render pass
- released action: 22 tests, 98.46% line coverage, 87.73% branch coverage, no production dependency vulnerabilities
- [fork PR #83](https://github.com/cpunion/llgo/pull/83) validated Linux/macOS publishing with no `main` baseline, then updated the same comment after a main baseline was added
- [generated #83 report](https://github.com/cpunion/llgo/pull/83#issuecomment-5102551991)
- [generated Pages series](https://cpunion.github.io/llgo-benchmark-data/llgo/baseline/?series=pull%2F83)
- [xgo-dev Pages dashboard](https://xgo-dev.github.io/llgo-benchmark-data/llgo/baseline/)
- [xgo-dev PR #2200 series](https://xgo-dev.github.io/llgo-benchmark-data/llgo/baseline/?series=pulls%2F2200)

The migration was also replayed against the real legacy main/#78 artifacts; all 38 legacy measurements and percentages matched before the obsolete TLS/GLS rows and serialized goroutine benchmark were corrected in response to review.


cpunion:
Addressed the review findings:

- documented the complete local benchmark/export commands and corrected ELF/Mach-O size terminology
- changed goroutine measurement to a bounded `sync.WaitGroup` batch
- fixed non-mutating true medians for even sample counts
- removed the now-unused project-specific report implementation, including its zero-baseline edge
- replaced the repeated custom publisher steps with the released reusable action workflow
- removed TLS/GLS measurements until their directives are available on `main`

Local package coverage is 92.1%; macOS collection and the released v1.0.1 record/render path pass.

PR #2201, opened 2026-07-28
cmd/compile: support type assertion diagnostics

MeteorsLiu:


cpunion (approved):

cpunion:
Closing because `interface/assertinline.go` is now explicitly classified on `main` as a cmd/compile-specific, not-applicable diagnostic case. LLGo does not need to reproduce this optimization diagnostic; the branch also predates the shared `internal/goflags` parsing path.

Issue #2154, opened 2026-07-22
proposal: add LLGo-aware LLDB integration


cpunion:
## Motivation

LLGo emits structurally valid DWARF, but current Apple and upstream LLVM LLDB builds do not maintain a Go language plugin. A compile unit marked `DW_LANG_Go` therefore loses otherwise valid parameters and locals in stock LLDB.

LLGo will not add or maintain a generic LLDB Go language implementation. Native debug compile units remain `DW_LANG_C`, which lets stock LLDB consume their standard DWARF locations, scopes, and types. LLGo identity and runtime presentation are provided separately.

## Goal

Provide an LLGo-aware LLDB integration that works with stock LLDB and real LLGo binaries:

- launch LLDB through an installed `llgo lldb` command;
- identify LLGo with `DW_AT_producer = "LLGo"` and a versioned debugger marker;
- preserve normal LLDB breakpoints, stepping, frames, parameters, locals, globals, and structural types;
- add marker-selected summaries, synthetic children, commands, and runtime views without replacing LLDB's ordinary variable model;
- keep the integration testable on macOS Apple LLDB and Linux LLVM LLDB.

## Compatibility contract

- LLGo debug compile units use `DW_LANG_C`.
- `DW_AT_producer = "LLGo"` remains the compiler identity used by standards-based DWARF validation.
- `__llgo_debugger_marker_v1` identifies an LLGo debugger ABI/runtime schema through LLDB's public API.
- Targets without the marker remain ordinary C/C++ targets; unsupported marker versions fail LLGo-specific presentation clearly without breaking raw debugging.
- The adapter must not depend on private LLDB APIs or a patched LLDB build.

The language tag is an interoperability choice only. LLGo must still emit correct DWARF types, scopes, locations, line tables, and producer metadata.

## Proposed structure

1. **Compiler and artifact contract**
   - emit standards-based DWARF consumable as `DW_LANG_C`;
   - retain the LLGo producer and versioned debugger marker;
   - describe pointer size, endianness, ABI version, and runtime-layout version.
2. **LLDB launcher and adapter**
   - install and load the LLGo Python adapter through `llgo lldb`;
   - use public `SB*` APIs and preserve direct LLDB use as a raw-debugging fallback;
   - select behavior only after validating the marker/schema version.
3. **Runtime presentation**
   - provide summaries and children for strings, slices, maps, channels, interfaces, function values, closures, and goroutines;
[...]

PR #2205, opened 2026-07-28
cmd: add LLGo-aware LLDB launcher

cpunion:
## Summary

Stock LLDB no longer has a maintained Go language plugin, and upstream requires
an RFC plus a committed long-term maintainer before restoring one. Keep the
LLGo-specific integration in this repository:

- add `llgo lldb`, which embeds and loads the LLGo Python adapter without
  depending on a source checkout;
- support LLDB 18+ auto-detection, `LLGO_LLDB`, `-lldb`, and LLDB argument
  pass-through;
- keep one plugin source and make the existing shell launcher a compatibility
  wrapper;
- keep adapter commands under the `llgo` namespace so stock LLDB `p` and `v`
  remain unchanged;
- keep the flat acceptance fixture under `cmd/llgo/lldbtest`, with source-line
  markers and explicit Python SB API assertions instead of comment parsing;
- run the complete LLDB acceptance suite through `llgo lldb` on the primary
  macOS and Ubuntu jobs.

This does not change LLGo's current DWARF contract: compile units remain
`DW_LANG_C` for stock-LLDB variable visibility, while the LLGo producer and
versioned marker select the external adapter. The test suite also verifies that
an ordinary C binary is not identified as LLGo.

## Testing

- macOS arm64, Apple LLDB 1703: 198/198 real breakpoint, scope, type, and value
  checks
- Ubuntu amd64, LLDB 19: 198/198 checks
- Linux arm64, LLDB 19.1.7: 198/198 checks
- `cmd/internal/lldb` statement coverage: 83.9% on both platforms
- `go test -vet=off -timeout 30m ./...`
- Linux: `go test ./cmd/llgo` and `go build ./...`

Refs #2154. Follow-ups #2206 and #2211.


cpunion:
Superseded by #2211, which retains the launcher and acceptance-suite commits and adds the versioned marker/schema contract. Keeping one standalone PR avoids an unnecessary merge dependency.

PR #2212, opened 2026-07-29
build: isolate invocation configuration and execution

zhouguangyuan0718:
## Summary

- initialize each LLGo command process `PATH` once, before builds start, from the existing `llvm.New("").BinDir()` discovery
- keep LLVM executables as process background instead of storing an LLVM installation in `Config`, build context, or backend workers
- introduce `Invocation` as the explicit `Build` input and keep `Do(args, conf)` as its compatibility wrapper
- clone caller-owned `Config` before resolving defaults and compatibility settings
- snapshot the invocation working directory and process environment once, then pass them explicitly to package loading and subprocesses
- restrict the internal `commandEnv` value to only child-process `Dir` and `Env`; cgo, pkg-config, llc, run, and emulator leaf helpers no longer receive the full compiler context only for those values
- move output-directory creation from configuration time to the actual link operation

## LLVM process model

`LLVM_CONFIG` is a process-startup choice. The shipped `cmd/llgo` entry point and repository tools that invoke `build.Do` call `llvm.SetupPath()` before starting builds. It prepends the selected bin directory using the platform path-list separator, ignores an empty directory, and does not duplicate an existing entry.

After `Build` starts, LLVM is not reselected or passed through build state. Existing tool usage remains ordinary:

- cgo probes execute `clang`
- IR checks execute `llc`
- archive selection uses `exec.LookPath("llvm-ar")`
- existing `llvm.Env` tool helpers retain their original behavior

Cross-compilation compilers and explicit overrides such as `LLGO_AR` retain their existing precedence.

## Invocation isolation

- `Invocation` contains only `Args`, `Config`, and `Dir`; it does not expose environment overrides
- `Config` slices/maps are cloned before defaults are resolved
- cwd and the process environment are captured once when `Build` starts
- `packages.Load` and child commands receive explicit `Dir`/`Env`
- relative outputs resolve against the invocation directory
- `NewDefaultConf` no longer creates `GOBIN`; the linker creates the selected output directory on demand
- no code under `build.Do`, `Build`, or package workers mutates process `PATH`
- `commandEnv` is deliberately limited to cwd/env and is not an extension point for compiler state

## Review size

- base: current `main` at `aa0bbe44`
- 18 files
- +659 / -80
- 1 commit

## Tests

- `go test ./internal/build ./cmd/llgo ./xtool/env/llvm -count=1`
- focused invocation, command environment, cgo probe, emulator, and output-directory tests
- the focused concurrency tests with `-race`
- `git diff --check`

### DIFF
--- a/chore/gentests/gentests.go
+++ b/chore/gentests/gentests.go
@@ -27,10 +27,12 @@ import (
 	"github.com/goplus/llgo/internal/littest"
 	"github.com/goplus/llgo/internal/llgen"
 	"github.com/goplus/llgo/internal/lto"
+	"github.com/goplus/llgo/xtool/env/llvm"
 	"github.com/goplus/mod"
 )
 
 func main() {
+	llvm.SetupPath()
 	dir, _, err := mod.FindGoMod(".")
 	check(err)
 
--- a/chore/litgen/litgen.go
+++ b/chore/litgen/litgen.go
@@ -23,9 +23,11 @@ import (
 	"path/filepath"
 
 	"github.com/goplus/llgo/internal/littest"
+	"github.com/goplus/llgo/xtool/env/llvm"
 )
 
 func main() {
+	llvm.SetupPath()
 	flag.Usage = func() {
 		fmt.Fprintf(flag.CommandLine.Output(), "Usage: %s <file-or-dir> [<file-or-dir>...]\n", filepath.Base(os.Args[0]))
 		flag.PrintDefaults()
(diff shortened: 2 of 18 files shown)


zhouguangyuan0718:
I reworked this prerequisite after narrowing the concurrency contract. The previous `processenv.Context`/request-Env implementation has been removed entirely.

The new head addresses the actual production side effect directly: LLVM tools are bound to `llvm-config --bindir`, and the build no longer mutates process `PATH`. Non-LLVM tools keep normal host semantics. This reduces this PR to 119 changed lines and avoids promising per-build PATH replacement that `go/packages` cannot fully honor.

PR #2213, opened 2026-07-29
ssa: terminate empty defer dispatch blocks

zhouguangyuan0718:
## Summary

- position the builder at the defer processing block when no deferred statements were registered
- keep the generated control-flow graph terminated in explicit defer-stack functions
- verify the regression case with LLVM module verification

This is the main-update compatibility fix extracted independently from #2196. It has no dependency on the process-isolation series.

## Size

- 2 files
- 11 additions, 1 deletion

## Tests

- `go test ./ssa -run ^TestExplicitDeferStackDrainWithoutLoopCases$ -count=1`
- `git diff --check`

### DIFF
--- a/ssa/eh.go
+++ b/ssa/eh.go
@@ -591,6 +591,9 @@ func (p Function) endDefer(b Builder) {
 			b.Jump(rethNext)
 		}
 	}
+	if n == 0 {
+		b.SetBlockEx(procBlk, AtEnd, true)
+	}
 	link := b.getField(b.Load(self.data), deferLink)
 	b.Call(b.Pkg.rtFunc("SetThreadDefer"), link)
 	b.IndirectJump(b.Load(rundPtr), nexts)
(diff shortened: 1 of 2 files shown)


zhouguangyuan0718:
Closing as an unproven/unreachable frontend case.

The regression test constructs the low-level builder state directly (`SetRecover` + `ssa:deferstack` + `RunDefers`) without any Go source defer or defer-stack drain. The real Go frontend preserves a stronger invariant:

- a normal source `defer` lowers through `Builder.Defer`, which appends an entry to `self.stmts`;
- a range-over-func defer makes `rangeFuncCallNeedsDeferDrain` insert `DeferStackDrain`, which also appends a drainer entry;
- without a source defer, x/tools removes the unused synthetic `ssa:deferstack`, and no relevant `RunDefers` path is emitted.

We were unable to produce a Go source fixture that reaches `endDefer` with `n == 0`. Main and the existing real-source rangefunc/defer suite remain green. The current verifier failure therefore demonstrates only an artificially assembled internal state, not a compiler bug reachable from Go. We can reopen with a baseline-failing Go-level reproducer if one is found later.

PR #2221, opened 2026-07-29
compiler: isolate per-invocation frontend state

zhouguangyuan0718:
## Summary

- pass frontend compilation options explicitly for each build invocation instead of changing package-level defaults
- capture debug info, tracing, export renaming, and shadow-stack behavior in invocation-owned `cl.Options`
- initialize LLVM frontend state once and keep compiler-verbose diagnostics build-local without frontend global switches or locks
- deprecate legacy frontend constructors that still read process-wide defaults
- verify concurrent invocations can compile with different DWARF/frontend settings under the race detector

## Dependency and review boundary

Based on current `main`, which includes #2246 and its unconditional main-package symbol prefix.

The intended review diff is:

- base: `8aa41210` (#2246 head)
- head: `c46521e4`
- 7 files
- +192 / -73
- 2 commits

Please review only `8aa41210..c46521e4`.

## Simplification from the previous version

- removed the `abi.Namer` and Program-owned naming policy
- removed per-Program `RewriteMainPrefix` state and its dual-policy concurrency tests
- removed scoped naming wrappers and named-type equivalence plumbing
- retained only frontend/compiler state that can genuinely vary between build invocations
- main package naming now follows #2246 everywhere and is not invocation state

## Compatibility boundary

- existing one-shot frontend constructors remain available, but are deprecated because they read legacy process defaults
- concurrent build drivers use `NewPackageExWithEmbedMetaOptions` with explicit `cl.Options`
- the build driver no longer calls legacy `cl.SetDebug` or `ssa.SetDebug`; compiler-verbose emits a package-level compile marker without process-wide state
- LLVM executable selection remains process-scoped as established by #2212

## Tests

- `GOCACHE=/private/tmp/llgo-2221-gocache go test ./internal/build ./cl ./ssa -count=1`
- `GOCACHE=/private/tmp/llgo-2221-gocache go test -race ./internal/build -run '^TestConcurrentInvocationsIsolateFrontendOptions$' -count=1`
- `GOCACHE=/private/tmp/llgo-2221-gocache go test ./cl -run '^TestFrontendOptions$' -count=1`
- focused atomic coverage: `legacyOptions` and `frontendOptions` both 100%; concurrent verbose/non-verbose invocation paths executed without a frontend lock
- `git diff --check`

### DIFF
--- a/cl/caller_frame_test.go
+++ b/cl/caller_frame_test.go
@@ -499,9 +499,7 @@ func TestRuntimeFrameNameNormalization(t *testing.T) {
 }
 
 func TestCompileRuntimeCallerFrameInstrumentation(t *testing.T) {
-	old := emitShadowStackInstrumentation
-	emitShadowStackInstrumentation = true
-	defer func() { emitShadowStackInstrumentation = old }()
+	t.Setenv("LLGO_SHADOW_STACK", "1")
 	ssapkg, files := buildCallerFrameSSAPackage(t, "example.com/foo", `package foo
 import "runtime/debug"
 
@@ -744,9 +742,7 @@ func top() {
 }
 
 func TestCompileRuntimeCallerFrameUsesGoNameForLinkname(t *testing.T) {
-	old := emitShadowStackInstrumentation
-	emitShadowStackInstrumentation = true
-	defer func() { emitShadowStackInstrumentation = old }()
+	t.Setenv("LLGO_SHADOW_STACK", "1")
 	ssapkg, files := buildCallerFrameSSAPackage(t, "command-line-arguments", `package main
 import "runtime"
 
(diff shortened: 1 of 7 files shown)


zhouguangyuan0718:
Reworked on top of #2246 (`7cbf71fe`) and force-updated to `ef96705d`.

The old SSA naming layer is gone: no `abi.Namer`, no Program-owned `RewriteMainPrefix`, no scoped naming wrappers, and no dual main-prefix race test. #2221 now contains only invocation-local frontend/compiler state.

I also addressed the remaining applicable review item by marking the legacy `NewPackage*` constructors deprecated, and replaced the old naming race test with concurrent builds using opposite DWARF/frontend settings. The new test passes under `-race`.

Validated locally:
- `go test ./internal/build ./cl ./ssa -count=1`
- `go test -race ./internal/build -run '^TestConcurrentInvocationsIsolateFrontendOptions$' -count=1`\n- `git diff --check`\n\nThe other prior review notes about scoped `Namer` methods and named-type equivalence no longer apply because those APIs were removed from this version.

zhouguangyuan0718:
CI follow-up at `8c821427`:

- Fixed `codecov/patch` by exercising the verbose frontend-debug lock path in the concurrent invocation test and adding focused coverage for explicit, legacy, and nil-context frontend options.
- Verified `TestFrontendOptions` passes, `TestConcurrentInvocationsIsolateFrontendOptions` passes under `-race`, and focused atomic coverage reports 100% for both `legacyOptions` and `frontendOptions`.
- The prior Ubuntu primary `hello` failure was an external HTTP 500 while downloading the ESP LLVM prebuilt archive; no source workaround was added. This head update has retriggered that job.

PR #2230, opened 2026-07-30
test/goroot: match vareq recovery diagnostics

cpunion:
Depends on #2222.

This narrowly scopes the extra parser-recovery diagnostics emitted for `test/syntax/vareq.go` to its exact primary diagnostic, source line, and location. Independent one-use groups preserve unrelated same-line diagnostics and do not consume duplicate follow-ons more than once.

Validation:
- `go test ./test/goroot -run "TestCheckExpectedErrorsScopesVareqRecovery|TestAdditionalParserRecoveryDiagnosticsFailOpen" -count=1`
- `go test ./test/goroot -count=1`
- `test/syntax/vareq.go` with Go 1.24.11, 1.25.0, 1.26.0, and 1.26.5
- remaining compile/errorcheck xfails with the same four GOROOT versions
- `git diff --check`

### DIFF
--- a/test/goroot/runner_test.go
+++ b/test/goroot/runner_test.go
@@ -1731,8 +1731,8 @@ func parserRecoverySourceSecondaries(primary, source string) []string {
 
 // parserRecoverySecondaryGroups returns independent one-use allowances for an
 // exact primary. Source-independent pairs are checked first, followed by
-// source-dependent single groups. issue11610 deterministically emits two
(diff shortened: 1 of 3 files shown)


cpunion:
Closing in favor of production frontend diagnostic normalization. The vareq recovery mapping is case-specific runner logic and should be replaced by the general loader strategy.

PR #2245, opened 2026-07-31
build: publish package archives concurrently

zhouguangyuan0718:
Stacked on #2243. Review range: `25c203a2..35b41702`.

## Summary

- write cacheable package archives directly to their final cache path, avoiding a temporary archive plus copy
- consume the #2175 LLVM `MemoryBuffer` members directly; no LLVM package object is written as a temporary `.o`
- preserve best-effort cache semantics by retrying an unwritable cache destination with a temporary `.a` only
- publish completed package artifacts with at most `min(-p, 2)` workers after all LLVM backends succeed
- keep cache warnings and fatal publication errors deterministic in package order
- remove build-owned temporary archives when safe, while retaining Darwin DWARF inputs required by later `dsymutil`
- emit package publication as `llgo.publish` worker spans in build traces

## Relationship to the archive change

The old publication benchmark was dominated by external `ar` rereading temporary LLVM objects. That path no longer exists, so those absolute timings and the previous backend/publish-overlap conclusion are not used as evidence for this rebased PR. Pipeline overlap can be reevaluated later; it is not required for the in-memory archive optimization.

## Validation

- `go test ./internal/build -count=1`
- focused `go test -race ./internal/build -run 'Test(NormalizeToArchive|PublishPackagePipelineTasks|CleanupTemporaryArchives)' -count=1`\n- direct-cache test uses a real LLVM bitcode MemoryBuffer\n- clean `-a -p=8` cgo + Plan9 build: 78 LLVM archives, 0 external `ar`, 0 temporary package `.a`, 0 temporary LLVM package `.o`; runtime output verified\n- `git diff --check`


zhouguangyuan0718:
Superseded by the restacked package worker design.

Archive publication is now part of each backend worker in #2182 (`8792dab9`), and #2242 pipelines that complete backend+archive+cache task directly (`c81529ce`). There is no longer a separate end-of-build publication sweep or publish worker pool, so this follow-up is no longer needed.

PR #2246, opened 2026-07-31
compiler: always use main symbol prefix

zhouguangyuan0718:
## Summary

- make packages named main always use the canonical main symbol prefix
- remove RewriteMainPrefix from build configuration, developer flags, llgen, and the global ABI switch
- keep synthesized test-main isolation by renaming the original main package to main.test
- refresh IR and metadata expectations for the now-unconditional naming behavior

This is an independent refactoring PR based directly on current main. It is not stacked on the package-parallel PR series.

## Validation

- go test ./internal/build ./cl ./ssa -count=1
- go test ./cl ./ssa -count=1 after all fixture refreshes
- go test -race ./ssa/abi -count=1
- go test -race ./internal/build -run ^TestFilterTestPackages$ -count=1
- go test ./ssa/abi ./cmd/internal/flags ./internal/llgen ./chore/llgen -count=1
- go test ./internal/build -run ^TestFilterTestPackages$ -count=1

### DIFF
--- a/chore/litgen/litgen_test.go
+++ b/chore/litgen/litgen_test.go
@@ -5,16 +5,10 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
-
-	"github.com/goplus/mod"
 )
 
 func TestProcessPath_SingleFileUsesContainingDir(t *testing.T) {
 	wd, err := os.Getwd()
-	if err != nil {
-		t.Fatal(err)
-	}
-	root, _, err := mod.FindGoMod(wd)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -26,7 +20,10 @@ func TestProcessPath_SingleFileUsesContainingDir(t *testing.T) {
 		_ = os.RemoveAll(pkgDir)
 	})
 	sourceFile := filepath.Join(pkgDir, "in.go")
-	if err := os.WriteFile(sourceFile, []byte("// LITTEST\npackage main\n\nfunc main() {}\n"), 0644); err != nil {
+	if err := os.WriteFile(sourceFile, []byte("// LITTEST\npackage main\n\nfunc main() { helper() }\n"), 0644); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(pkgDir, "helper.go"), []byte("package main\n\nfunc helper() {}\n"), 0644); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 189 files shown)


zhouguangyuan0718:
Addressed the review and CI failures in 0fb90ad3:

- removed the now-unreachable main-prefix rewrite from funcInfoDisplayName and simplified its call/test
- updated the litgen single-file test while preserving its containing-package coverage with a sibling helper
- refreshed the remaining dev-LTO main symbol checks
- changed internal/cabi ABI fixtures to look up main.* symbols and fail cleanly on a missing function instead of calling the LLVM C API with a nil value

Local validation:
- go test ./chore/litgen -count=1
- go test ./internal/cabi -count=1
- go test -tags=dev ./cl -run ^TestRunAndTestFromTestlto/globaldce_ -count=1
- go test ./cl -run ^TestFuncInfoDisplayName$ -count=1

zhouguangyuan0718:
Follow-up for the rerun of **Dev LTO GlobalDCE**:

The refreshed IR goldens passed, but the later `nm`/FileCheck layer still expected the old per-fixture main-package prefix. Commit `cf5c5825` updates all 17 main-package `SYMBOL-*` expectations to `main`, while preserving imported subpackage paths.

Validation:
- `go test -tags=dev ./cl -run '^TestBuildAndCheckSymbolsFromTestlto' -count=1`\n- `git diff --check`\n\nThe LLVM 19 pass-plugin link path cannot be exercised on Darwin (`ld64.lld` does not support LTO pass plugins), so the plugin subset is being revalidated by the Linux CI rerun.

zhouguangyuan0718:
Ubuntu follow-up (commit `7cbf71fe`):

The remaining Linux-only symbol failures were caused by PCLN/funcinfo linking with `-Wl,--export-dynamic-symbol=main.*`. Once all main packages use the canonical prefix, that export policy intentionally retains every `main.*` symbol and masks GlobalDCE in the final symbol-table tests. The tests now set `PCLNMode = PCLNNone` for symbol-only checks, so they continue to assert dead-symbol removal without changing production linking behavior or weakening `SYMBOL-NOT` coverage.

Verified on Ubuntu with LLVM 19 and the built `LLGOLTOPlugin.so`:
- all non-plugin GlobalDCE symbol fixtures: PASS (95.9s)
- all plugin symbol fixtures: PASS (60.8s)
- plugin run + symbol suites together: PASS (132.8s)

The current CI rerun is testing the same head and Linux path.

PR #2252, opened 2026-08-01
runtime: preserve type metadata in assertion panics

cpunion:
Type assertion failures currently lower the asserted type to a string before calling the runtime. That loses package identity, the source interface type, and instantiated generic type information.

Pass ABI type descriptors for the source interface and asserted type instead, and construct the existing Go-compatible `TypeAssertionError` directly. Empty source interfaces remain nil so ordinary interface-conversion messages keep their Go spelling.

This removes the xfails for:
- `fixedbugs/issue18911.go`
- `interface/embed3.go`
- `typeparam/mdempsky/16.go`
- `fixedbugs/issue26094.go`

The scoped same-name type regression is covered in `TestInterfaceAssertRejectsSameNameTypesFromDifferentScopes`; it checks both local-to-local and local-to-package assertions.

Validation:
- `go test ./ssa ./test/goroot -count=1 -timeout=15m`
- affected 14 cl runtime/IR fixtures
- focused llgo test assertion regressions
- the first three GOROOT cases with Go 1.24.11, 1.25.0, and 1.26.5 on darwin/arm64
- `fixedbugs/issue26094.go` with empty xfail on all three Go versions on darwin/arm64

### DIFF
--- a/cl/_testdata/vargs/in.go
+++ b/cl/_testdata/vargs/in.go
@@ -88,7 +88,7 @@ func test(a ...any) {
 // CHECK-NEXT:   br label %_llgo_1
 // CHECK-EMPTY:
 // CHECK-NEXT: _llgo_5:                                          ; preds = %_llgo_2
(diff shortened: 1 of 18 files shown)


cpunion:
Superseded by #2034. The type-assertion metadata implementation and its GOROOT conformance cases are preserved there as their own layered commit.

PR #2253, opened 2026-08-01
ssa: saturate float-to-uint32 conversions

cpunion:
Go 1.26 `convert5.go` selects the new saturating float-to-uint32 behavior with `-gcflags=-d=converthash=qy`. LLGo previously ignored that frontend flag and always used its legacy int64-then-truncate path.

Parse cmd/compile `-d` settings generically in `internal/goflags`, with Go-compatible last-value-wins behavior. Global hash-debug patterns share the `internal/bisect` syntax, so `q`, repeated `v`, `!`, and the `n`/`y` aliases are handled independently of `converthash`; the same parser can be reused by `fmahash`, `loopvarhash`, and other hash-debug settings. LLGo maps only patterns that enable every candidate to its build-wide option, while selective hash patterns remain forwarded to `go/packages`.

Carry the typed setting through the build configuration and SSA target, and include it in the build cache fingerprint. The default conversion path remains unchanged, so existing programs and demos keep their target-compatible legacy results; only explicitly flagged builds use the guarded unsigned conversion that clamps negative and out-of-range values.

This removes the Go 1.26 macOS/arm64 and Linux/amd64 xfails for `convert5.go`.

Validation:
- GOROOT `convert5.go` with Go 1.26.5 and an empty xfail file on macOS/arm64 and Linux/amd64
- review counterexamples `_demo/go/issue1538` and `_demo/go/issue1538-floatcvtuint-over` in default mode on both platforms
- `go test ./internal/goflags`
- focused build fingerprint and legacy/saturating IR checks
- full `go test ./ssa`
- Go 1.26 compiler acceptance checked for the supported global bisect-pattern spellings

The initial unconditional implementation failed the two `issue1538` demos. This version deliberately preserves their default behavior and enables saturation only for an applicable explicit compiler flag.

### DIFF
--- a/cl/float_int_conversion_compile_test.go
+++ b/cl/float_int_conversion_compile_test.go
@@ -1,0 +1,44 @@
+//go:build !llgo
+
+package cl
+
+import (
+	"runtime"
+	"strings"
+	"testing"
+
+	llssa "github.com/goplus/llgo/ssa"
+)
+
+func TestFloatToUint32ConversionMode(t *testing.T) {
+	const src = `package floatconvert
+func Convert(x float32) uint32 { return uint32(x) }
+`
+	tests := []struct {
+		name       string
+		saturating bool
+		want       string
+		notWant    string
+	}{
+		{name: "legacy", want: "fptosi float", notWant: "fptoui float"},
+		{name: "saturating", saturating: true, want: "fptoui float", notWant: "fptosi float"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ssaPkg, _, files := buildGoSSAPkg(t, src)
+			prog := newLLSSAProgForTarget(t, &llssa.Target{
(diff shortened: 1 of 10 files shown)


cpunion:
@fennoai Review addressed after rebasing onto the latest main.

- The two stale-demo counterexamples are preserved intentionally: LLGo now keeps the legacy float-to-uint32 path by default and enables saturation only for an applicable explicit `-d=converthash` pattern. Both `_demo/go/issue1538` and `_demo/go/issue1538-floatcvtuint-over` pass unchanged on macOS/arm64 and Linux/amd64.
- Added the suggested comment explaining that `dstSize < 4` retains the existing signed-convert-and-truncate behavior because Go's `converthash` transition targets float-to-uint32.
- Generalized `-d` parsing and the shared Go bisect-pattern modifiers instead of hard-coding `y`/`qy` for `converthash`.

GOROOT `convert5.go` passes with an empty xfail file on both platforms.

PR #2282, opened 2026-08-04
build: prepare read-only package backend state

zhouguangyuan0718:
## Summary

- collect linknames, exports, closure-env directives, nointerface markers, type backgrounds, package syntax identities, and locality metadata before backend lowering starts
- create each backend Program with a fresh LLVM Context, TargetMachine, lowering/type caches, ABI state, and C ABI transformer while directly sharing the already-prepared Go metadata read-only
- precompute caller-frame tracking for every Go SSA package before workers start
- preload the effective patched package syntax explicitly because patch.Types has a distinct go/types identity

## Design boundary

This PR is stacked on #2280. It deliberately replaces the earlier snapshot/overlay approach: there is no freeze API, syntax base/delta, worker merge, syntax validation pass, or metadata copying. One-shot compiler users retain Program-local mutation; the build driver switches to read-only syntax mode only after all package preparation is complete.

Parallel scheduling remains in the following PR; this change only establishes the minimal safe state boundary it needs.

## Validation

- go test ./ssa ./cl ./internal/build -count=1
- go test ./ssa ./cl ./internal/build -coverprofile=... -count=1
- targeted race tests for backend Program sharing, preloaded syntax, caller tracking, and build sessions
- new executable-line patch coverage measured locally at about 92.3%

### DIFF
--- a/cl/caller_tracking_precompute_test.go
+++ b/cl/caller_tracking_precompute_test.go
@@ -1,0 +1,61 @@
+//go:build !llgo
+
+/*
+ * Copyright (c) 2026 The XGo Authors (xgo.dev). All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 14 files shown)


zhouguangyuan0718:
@xgopilot please review the current PR head (c1728af80cbb41408b54cb1b1b94e4902edc36ff) from scratch. Focus on whether package syntax, locality, caller-tracking, runtime/Python package data, and frontend options are fully prepared before backend workers start; whether shared state is truly read-only; whether worker-local writes can leak or race; patched/alternate-package behavior; order independence and deterministic output; and regression-test completeness. Please ignore findings that only apply to earlier revisions.

zhouguangyuan0718:
@xgopilot review current head c1728af80cbb41408b54cb1b1b94e4902edc36ff. Focus on freeze-before-share correctness, worker read-only state, patched/alternate packages, races, determinism, and missing tests.

PR #2455, opened 2026-08-30
cabi: prove parameter storage reuse

zhouguangyuan0718:
## Summary

- replace the indirect-parameter alloca-count heuristic with proof-based parameter-home reuse
- reuse at most one entry-block home after checking allocation shape, initialization dominance, pre-initialization uses, and alignment
- apply the same policy to pointer-, one-register-, and two-register-lowered aggregate parameters
- materialize width-lowered returns from the already evaluated SSA value instead of re-reading a load source
- cover independent copies, repeated stores, memset prefixes, escapes, conditional initialization, stronger alignment, and small aggregate returns

## Root cause

`github.com/Masterminds/semver/v3` exposed this when `Version.SetPrerelease` and `Version.SetMetadata` copied a 72-byte value receiver, mutated the copy, and later read the original receiver. The old replacement walked every `store param -> alloca` and redirected every destination to the same ABI storage, collapsing independent objects into one address. A `len(allocas) == 1` guard prevented the reproducer but encoded the symptom rather than the substitution contract.

The new path treats ABI storage as a replacement for at most one proven parameter home. If dominance, pre-initialization aliasing, allocation shape, or alignment cannot be established, it leaves that alloca intact.

## Testing

- `go test ./internal/cabi -count=1`
- `go test ./ssa ./cl -count=1`
- `go test ./internal/build -run ^TestLargeArrayReturnAllABIModes$ -count=1`
- `llgo test -v -count=1 -run ^(TestSetPrerelease|TestSetMetadata)$ .` in `github.com/Masterminds/semver/v3@v3.3.1`

The ABI runtime regression passes in modes 0, 1, and 2. The original semver failures also pass with the compiler built from this head.

### DIFF
--- a/internal/build/large_array_return_test.go
+++ b/internal/build/large_array_return_test.go
@@ -24,7 +24,7 @@ func TestLargeArrayReturn(t *testing.T) {
 	if len(lines) < 2 {
 		t.Fatalf("output has fewer than two lines: %q", lines)
 	}
-	if got, want := strings.Join(lines[len(lines)-2:], "\n"), "99 0 0 0 98 98 6\n97"; got != want {
+	if got, want := strings.Join(lines[len(lines)-2:], "\n"), "99 0 0 0 98 98 6 1\n97"; got != want {
 		t.Fatalf("output = %q, want %q", got, want)
 	}
 }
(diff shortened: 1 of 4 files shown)


visualfc (approved):
commit 有重复，可以简化

PR #2497, opened 2026-09-04
[WIP] cl, runtime, reflect: support typed SetFinalizer and noffi method values

visualfc:
## Summary

This PR adds compiler and runtime support for statically typed `runtime.SetFinalizer` calls and a noffi reflect mode for builds that do not need libffi.

## Implementation

- Lower compatible statically typed `runtime.SetFinalizer` calls to `runtime.SetFinalizerPtr`.
- Generate an internal typed wrapper that converts the runtime `unsafe.Pointer` back to the finalizer's pointer argument, including `any` and implemented interface parameters via `MakeInterface`.
- Preserve the original execution timing for direct calls, `defer`, and `go` calls.
- Lower `runtime.SetFinalizer(obj, nil)` to the pointer-based cancellation path.
- Keep the boxed `runtime.SetFinalizer` path for capturing closures and for runtimes that do not provide `SetFinalizerPtr`, including WASM and collector-free runtimes.
- Add pointer-based finalizer registration, replacement, cancellation, callback queuing, and execution handling.
- Preserve cleanup chains installed before or after a typed finalizer when that finalizer is canceled or replaced.
- When `-check-libffi=true`, scan the already-built Go SSA with RTA from `init`/`main` (or every function for libraries and tests). Restart once with `llgo_noffi` and `llgo_methodvalue_noffi` if `reflect.Value.Call`, `CallSlice`, `MakeFunc`, Windows `NewCallback`, or an unlowered `SetFinalizer` are unreachable. LLVM backend and LTO still run once.
- Skip that scan on WASI Preview 1 (`wasip1`, `-target wasi` / `wasip1`). WASI has no host libffi and uses typed reflection bridges.
- On js/wasm, pass `-flto` through to `clang`/`emcc -c` so `-lto=full` emits bitcode and can DCE unused `$methodvalue` thunks.
- On js/wasm, noffi builds omit the libffi Asyncify bind removelist (`llgo_reflect_bind*_js`) so emcc does not warn about unmatched patterns.
- Use ABI-correct `$methodvalue` thunks for noffi `MethodByName().Interface()` values so the receiver is passed as an ordinary method argument.
- Keep `reflect.MakeFunc` and `reflect.Value.Call` unavailable when `llgo_noffi` is enabled.
- Keep Windows `syscall.NewCallback` and `syscall.NewCallbackCDecl` unavailable in noffi mode.

## Build Flag

`llgo build -check-libffi=true` opts into automatic libffi selection. It is off by default, including with `-lto=full`. Eligible builds scan Go SSA reachability and compile once without libffi when the package graph does not need it. WASI Preview 1 ignores the flag.

## Comparison

Program: `cmd/_demo/hellow` (`fmt.Println("hello world")`). Measured on macOS arm64 against current `xgo-dev/main` (`81a4d030f`) and this PR (`971590ba8` plus local follow-ups).

js/wasm32 rows use `-lto=full` on this PR. Current `main` does not pass `-flto` to wasm `clang -c`, so those compiles do not complete here; the wasm comparison is this PR with libffi versus this PR with `-check-libffi`.

| Platform | Build | Flags | Size | Change | Links `libffi` |
|---|---|---|---:|---:|:---:|
| macOS arm64 | `main` | `-lto=full` | 926992 bytes | - | dynamic link |
| macOS arm64 | This PR | `-lto=full -check-libffi` | 926688 bytes | -304 B / -0.03% | No |
| js/wasm32 | This PR | `-lto=full` | 1205032 bytes | - | static |
| js/wasm32 | This PR | `-lto=full -check-libffi` | 1139798 bytes | -65234 B / -5.4% | No |


ssa/expr.go
```diff
@@ -2125,4 +2148,80 @@ func llvmFields(vals []Expr, t *types.Struct, b Builder) (ret []llvm.Value) {
...
+	name := fmt.Sprintf("%s.$setFinalizer$%s$%s$%s",
+		b.Pkg.Path(), direct.Name(), objName, b.Prog.abi.FuncName(finalizerSig))
+	wrapper := b.Pkg.FuncOf(name)
+	if wrapper == nil {
+		wrapper = b.Pkg.NewFunc(name, wrapperSig, InGo)
+		wrapper.impl.SetLinkage(llvm.InternalLinkage)
+		wb := wrapper.MakeBody(1)
+		converted := wb.Convert(b.Prog.rawType(ptrType), wb.Param(0))
+		if !types.Identical(argType, ptrType) {
+			converted = wb.Convert(b.Prog.rawType(argType), converted)
```
cpunion: Blocking: setFinalizerArgCompatible explicitly accepts empty and implemented interfaces, but this is a pointer-to-interface conversion and Builder.Convert does not construct an interface value. A named finalizer such as func finalize(v interface{ Value() int }) passed to runtime.SetFinalizer(p, finalize) reliably aborts compilation with package job ... panicked: todo; named func(any) has the same problem. The existing assignability tests use capturing closures, so they remain on the boxed/libffi path and never exercise this wrapper. Please use the proper interface-boxing operation here and add named-function compile-and-run cases for both any and a non-empty implemented interface.

ssa/expr.go
```diff
@@ -1608,6 +1609,28 @@ const (
 	ReflectMethodMask     = ReflectMethodByIndex | ReflectMethodByName | ReflectMethodDynamic | reflectTypeMethodMask
 )
 
+func (b Builder) checkFFI(fn Expr) {
+	pkg := b.Pkg
+	if !pkg.NeedFFI {
+		switch fn.Name() {
+		case "reflect.Value.Call", "reflect.Value.CallSlice", "reflect.MakeFunc":
```
cpunion: Blocking: this detection only observes direct vkFuncDecl calls. A higher-order call such as invokeMakeFunc(reflect.MakeFunc), where invokeMakeFunc accepts the function as a value and calls it, never reaches checkFFI because the vkClosure path returns earlier. With -lto=full, automatic selection then omits libffi and the resulting program panics with reflect.MakeFunc is unavailable without libffi. Any value/address escape of an FFI-sensitive symbol must conservatively set NeedFFI, or the decision must use real post-lowering reachability. Please add this exact higher-order case as a regression test.

runtime/internal/lib/reflect/methodvalue_noffi.go
```diff
@@ -0,0 +1,53 @@
...
+	_, _, fn := methodReceiver(op, rcvr, int(v.flag)>>flagMethodShift)
+	var receiver unsafe.Pointer
+	storeRcvr(v, unsafe.Pointer(&receiver))
+	env := &directMethodValueEnv{
+		marker: &directMethodValueMarker, receiver: receiver, receiverType: rcvr.typ(),
+	}
+	fv := &struct {
+		fn  unsafe.Pointer
+		env unsafe.Pointer
+	}{fn, unsafe.Pointer(env)}
```
cpunion, line 37: Blocking: fn is the real method entry, which expects the receiver as its ordinary first ABI argument, while env is supplied as the hidden closure context. Calling reflect.ValueOf(&T{}).MethodByName("M").Interface().(func(int) int)(2) does not pass through reflect.Value.call to decode directMethodValueEnv; it jumps to fn with the wrong ABI. I reproduced a nil-pointer panic with automatic noffi, while the same program built with -check-libffi=false correctly prints 42. This representation needs an ABI-correct thunk that extracts the receiver from env and calls the method with it as the ordinary receiver argument.

runtime/internal/runtime/z_gc.go
```diff
@@ -201,6 +290,67 @@ func AddCleanupPtr(ptr unsafe.Pointer, cleanup func()) (cancel func()) {
...
+func cancelPointerFinalizer(obj unsafe.Pointer, old *pointerFinalizerEntry) {
+	if old != nil {
+		// Stop a leftover boxed registration without restoring it: old.prevFn
+		// is the cleanup chain, or a boxed callback that now chains to it.
+		if CancelBoxedFinalizer != nil {
+			CancelBoxedFinalizer(obj, false)
+		}
+		var ignoredFn bdwgc.FinalizerFunc
+		var ignoredCb unsafe.Pointer
+		bdwgc.RegisterFinalizer(obj, old.prevFn, old.prevCb, &ignoredFn, &ignoredCb)
```
cpunion: Blocking: this restores the callback snapshot captured when the typed finalizer was registered. If runtime.AddCleanup installs a newer callback chain afterward, restoring old.prevFn overwrites and discards that chain. For SetFinalizer(p, typedFinalizer); AddCleanup(p, cleanup, arg); SetFinalizer(p, nil), official Go reports cleanups 8 old 0, while this head consistently reports cleanups 0 old 0. Cancellation must preserve callbacks installed after the typed entry, and this ordering needs a runtime regression test.

internal/build/build.go
```diff
@@ -871,6 +873,16 @@ func Build(inv Invocation) (result []Package, resultErr error) {
...
 		return nil, err
 	}
+	// When enabled, retry once with the noffi reflect implementation if the
+	// complete package graph does not require libffi.
+	if conf.CheckFFI && !conf.noFFIRestart && !ctx.needFFI && hasLinkedReflect(allPkgs) && !hasBuildTag(conf.Tags, "llgo_noffi") {
+		restarted := conf.clone()
+		restarted.Tags = appendBuildTag(restarted.Tags, "llgo_noffi")
+		restarted.Tags = appendBuildTag(restarted.Tags, "llgo_methodvalue_noffi")
+		restarted.noFFIRestart = true
+		return Build(Invocation{Args: inv.Args, Config: restarted, Dir: dir,
```
cpunion, line 885: This recursively rebuilds the complete package graph after the first build has already finished. The benchmark attached to this PR shows full-LTO fmtprintf build-time regressions of +23.0% on Linux, +40.0% on macOS, and roughly +30% to +35% on Windows, for about 1.4% to 3.2% file-size reduction. Please select the mode before the expensive backend work or reuse the first-pass results; at minimum this cost should be documented explicitly and the default-on policy reconsidered.

cpunion (commented):
I verified that the size-reduction direction is real, but the current implementation is not safe to merge yet. The inline comments contain four deterministic correctness failures reproduced against this head and one material build-time issue. The current tests miss the named-interface-finalizer and higher-order MakeFunc paths. This branch is also 42 commits behind current main and overlaps the recent runtime cleanup/reentrancy work, so it needs a rebase and full retest after the correctness fixes. I recommend separating typed SetFinalizer lowering/runtime support from automatic noffi reflect/method-value selection if that makes the invariants and review surface easier to establish.

ssa/expr.go
visualfc: Fixed. Interface arguments now go through `MakeInterface` instead of `Convert`. Named `func(any)` and `func(interface{ Value() int })` compile-and-run cases are in `TestRuntimeSetFinalizerNamedInterfaceFunctions`.

ssa/expr.go
visualfc: Fixed. `checkFFI` now also runs when an FFI-sensitive function escapes as a value (`vkFuncDecl` → closure in `checkExpr`, and `MakeClosure`). The higher-order case `invokeMakeFunc(reflect.MakeFunc)` is covered by `TestMakeFuncValueRequiresFFI` / `TestCheckFFIMarksHigherOrderMakeFunc`.

runtime/internal/lib/reflect/methodvalue_noffi.go
visualfc: Fixed. Each method now gets an ABI-correct `$methodvalue` thunk that loads the receiver from the hidden env and calls the real method with it as the ordinary first argument. Thunk pointers live in a side array after the method table (`abi.Method` is unchanged). `MethodByName("M").Interface().(func(int) int)(2)` is covered by `TestReflectValueMethodByNameInterfaceCall`.

runtime/internal/runtime/z_gc.go
visualfc: Fixed. Cancellation now probes the current BDWGC slot: if a later `AddCleanup` chain owns it, that chain is put back instead of restoring the registration-time snapshot. Stopped typed entries also forward `prevFn` so earlier cleanups are not dropped. Regression: `TestRuntimeSetFinalizerNamedCancelAfterCleanup` (`SetFinalizer(named)` → `AddCleanup` → `SetFinalizer(nil)`).

internal/build/build.go
visualfc: Fixed. CheckFFI now does a frontend-only scan of ordinary packages and, if libffi is not needed, reloads with `llgo_noffi` before any package LLVM backend or LTO. `-lto=full` no longer enables this by default; it is opt-in via `-check-libffi=true`.

PR #2539, opened 2026-09-08
wasm: Emscripten browser fs host without fs.*Sync overlay

visualfc:
## Summary

On Emscripten (`-target emscripten` / `emscripten-memory64` / `wasm`), Go `syscall/fs_js.go` talks to Node-shaped `globalThis.fs`. The browser has no Node `fs`, and Emscripten `Module.FS` is a different API.

This PR ships `targets/wasm_fs.js` as that browser host and copies it next to Emscripten JS/HTML output. It does **not** replace official `fsCall` with `fs.*Sync`: after rebase, `js.FuncOf` is synchronous and the callback+channel path works.

- Keep official `syscall/fs_js.go` (no overlay).
- `wasm_fs.js` only for named Emscripten targets. GoJS keeps `wasm_exec.js` / Node `fs`.
- Install the `js.FuncOf` invoke bridge once so reinstalling it cannot drop queued host events.
- Honor Emscripten `-o` suffixes (`.js` / `.mjs` / `.html` / `.wasm`). A `.wasm` output still links a sibling `.mjs`.
- Patch `TextDecoder` and `crypto.getRandomValues` so Chrome can use growable `WebAssembly.Memory`.

## Install at build time

For executables with `-target emscripten`, `emscripten-memory64`, or `wasm` whose driver output is `.html`, `.js`, or `.mjs`, `llgo build` copies `LLGO_ROOT/targets/wasm_fs.js` next to the output:

- `-o main.html`: write `wasm_fs.js` beside the page and insert a script tag for `./wasm_fs.js` immediately before the first `type=module` script
- `-o main.js` / `-o main.mjs`: copy `wasm_fs.js` only; a custom HTML page must load it first
- `-o main.wasm`: emcc still emits `main.mjs` + `main.wasm`; the shim is copied next to the `.mjs`

Node tests keep using native `fs` (`emscripten-node-polyfills.mjs`) and do not load this shim.

## Demo

```bash
cat > main.go <<'GO'
package main

import (
	"fmt"
	"os"
)

func main() {
	println("hello")
	fmt.Println("hello world")
	os.Stdout.Write([]byte("hello2\n"))
}
GO

llgo build -target emscripten -o main.html .
# Serve the directory over HTTP (not file://) and open main.html.
```

A custom page should bind the shim to the Emscripten module **before** Go runs: load `./wasm_fs.js`, call `globalThis.llgoAttachWasmFS(Module)`, then `await initModule(Module)`.

## Test plan

- [x] `go test ./internal/build` host/output tests (`TestNeedsEmscriptenBrowserHost`, `TestWasmFSHostShim`, `TestSyscallSourcePatch`, ...)
- [x] `llgo test -target emscripten -emulator ./internal/build/testdata/wasm-test` (`TestJSCallFileIO`, `TestFSCallAsyncFallback`, `TestSyncFileIO`)
- [x] Browser: `llgo build -target emscripten -o main.html .` and open the page (`fmt.Println` + `os.Stdout`)


runtime/_patch/syscall/fs_js.go
```diff
@@ -0,0 +1,48 @@
...
+import "syscall/js"
+
+// fsCall prefers Node's fs.*Sync APIs so file I/O does not park on a js.FuncOf
+// callback. Official Go waits on a buffered channel after fs.write; that
+// nested park aborts LLGo's Emscripten Fiber/Asyncify scheduler. Methods
+// without a Sync counterpart keep the original callback-and-channel path.
+func fsCall(name string, args ...any) (val js.Value, err error) {
+	syncName := name + "Sync"
+	if !jsFS.Get(syncName).IsUndefined() {
+		defer recoverErr(&err)
```
cpunion, line 14: [P1] `recoverErr` cannot currently recover an exception thrown by this synchronous JS call. I built the exact PR head and ran `os.Stat("/llgo-pr2539-definitely-does-not-exist")`: Node throws an uncaught `ENOENT` and the process exits 1 instead of returning an error for which `os.IsNotExist` is true. In the same binary, setting `fs.statSync` to `undefined` to select the old async path returns `ENOENT` normally and exits 0. The emval call bridge must preserve/raise the JS error as `js.Error`, or this path needs a JS wrapper that returns the error object; please add an end-to-end failing-filesystem-call test. The browser shim also needs to normalize direct `FS.write`/`FS.chdir` exceptions because optimized Emscripten `ErrnoError` objects have `errno` but no Node-style `code`.

runtime/_patch/syscall/fs_js.go
```diff
@@ -0,0 +1,48 @@
...
+		return jsFS.Call(syncName, args...), nil
```
cpunion, line 15: [P2] Selecting every available Sync method makes blocking stdin freeze the entire single-threaded Go/JS scheduler. In an end-to-end test, one goroutine scheduled a 20 ms timer and the main goroutine called `os.Stdin.Read`; input arrived after 300 ms. With this path the timer fired only after the read returned (~372 ms), while disabling `fs.readSync` let it fire at ~71 ms while the read remained pending until ~352 ms. Interactive programs therefore cannot run timers, goroutines, or JS callbacks while waiting for input. At least potentially blocking stream operations such as `read` need to retain the async path rather than choosing solely by method existence.

internal/build/build.go
```diff
@@ -2227,15 +2227,22 @@ func executeMainLink(ctx *context, plan *mainLinkPlan, verbose bool) error {
...
-	defer cleanupWasmLinkOutput(linkOutput, plan.outputPath)
+	defer cleanupWasmLinkOutput(linkOutput, driverOut)
 	if err := linkObjFiles(ctx, linkOutput, plan.linkInputs, plan.linkArgs, verbose); err != nil {
 		return err
 	}
-	return publishWasmLinkOutput(ctx, linkOutput, plan.outputPath, verbose)
+	if err := publishWasmLinkOutput(ctx, linkOutput, driverOut, verbose); err != nil {
+		return err
+	}
+	if err := removeStaleEmscriptenGlue(driverOut); err != nil {
```
cpunion, line 2242: [P1] This cleanup runs after every executable link, not only Emscripten links. I reproduced a native `llgo build -o app.js main.go` deleting a pre-existing unrelated sibling `app.mjs` while successfully producing `app.js`. Please gate stale-glue cleanup on the Emscripten target/output contract, and ideally remove only sidecars that this build mode is known to own.

internal/build/emscripten_html.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+	if ctx.shouldPrintCommands(verbose) {
+		fmt.Fprintf(os.Stderr, "copy %s -> %s\n", src, filepath.Join(filepath.Dir(output), wasmFSScriptName))
+	}
+	return nil
+}
+
+func installEmscriptenBrowserHost(src, output string) error {
+	dst := filepath.Join(filepath.Dir(output), wasmFSScriptName)
```
cpunion, line 101: [P2] The fixed sidecar name can alias the requested output. I reproduced `llgo build -target emscripten -o wasm_fs.js main.go` succeeding, then this copy replaced the generated module with the shim itself (the final file was byte-for-byte `targets/wasm_fs.js` and no longer exported an Emscripten module factory). Please reject this collision or choose a non-colliding output-specific sidecar name, with a regression test.

cpunion (commented):
I reviewed the current sync-first implementation and output publication paths with actual LLGo/Emscripten binaries. The happy-path tests pass, but the four inline findings below reproduce correctness regressions that should be fixed before merge.

PR #2572, opened 2026-09-12
refactor: keep existing function directives on declaration objects

zhouguangyuan0718:
Existing function directives are stored in separate program-wide tables and looked up again during lowering. This refactor gives each source function or method a declaration object and carries that object with the frontend function.

The declaration owns the existing `//llgo:env`, `//go:wasmimport`, `//go:nointerface`, function linkname and export information. Parsing creates the object first, then sets its properties. The dedicated env, wasm-import and nointerface tables are removed. Function links and exports move out of the shared tables when their declarations are bound; variable and unresolved-symbol entries retain their existing lookup paths.

The existing frontend-to-backend association now stores an `aFunction` wrapper containing the Go SSA function, its source declaration and its backend entry. Declaration identity includes the package and source location. Two declarations that link to the same LLVM symbol keep independent metadata, while the existing backend function table still resolves symbols and checks closure-environment ABI compatibility. Function constructors keep their existing signatures.

Generic instances retain their origin declaration, including private functions omitted from export-only preload. Both preloaded builds and the one-shot compilation entry point explicitly bind package-patch replacement declarations; original declarations retain their own information. The one-shot path selects replacements from the alternate SSA package, including declared methods, because its source files also contain the original declarations. Prepared declarations contain no LLVM state and are shared read-only by backend Programs.

Validation on macOS arm64 with Go 1.27.0 and LLVM 22.1.8:

- Complete `cl`, `ssa`, and `internal/build` suites rerun after the one-shot patch-binding fix.
- Complete `cl` suite with `LLGO_BUILD_CACHE=off`, plus complete `ssa` and `internal/build` suites. These include existing env ABI, wasm import, nointerface, export and locality/linkname coverage.
- Execution tests for ordinary builds and full LTO with caching enabled after invalidating stale development artifacts.
- Regression tests for independent aliases and package variants, pending method declarations, generic origin binding after export-only preload, patched imported intrinsics, and one-shot patch binding for ordinary functions, generic functions, and value/pointer receiver methods.
- Runtime tests at `-O2` with default GC and `nogc`.
- Cross-package cache checks with a generic caller: first build, cache hit, linkname-only retargeting, restoration and another cache hit. Retargeting recompiles the library and changes the executable result; restoration reverses it.


zhouguangyuan0718:
@fennoai please review the current PR head, 1df08aa4064038a26f8d5dda9c6e539e67b1720e, against main. The PR now only refactors existing function declaration metadata (env, wasmimport, nointerface, function linknames and exports); it no longer introduces source contract attributes. Please focus on correctness across preloaded and one-shot compilation, explicit package patches, generic origins, imported functions, and independent source declarations that share a linker symbol. Report reproducible regressions and missing tests without modifying code.

PR #2600, opened 2026-09-15
build: preserve native test termination signals

cpunion:
## Problem

`llgo test` currently reports a native test process killed by a signal only as `exit code -1`. This loses the operating-system reason needed to distinguish a segmentation fault, profiling signal, or external kill. The missing diagnostic obstructs investigation of [the macOS native test/go failure in #2578](https://github.com/xgo-dev/llgo/actions/runs/34899355143/job/104161289191). This change does not claim to fix or identify the cause of that crash.

## Change

Use `ExitError.Exited()` to distinguish an ordinary process exit from signal termination. Preserve the existing `exit code N` format for normal exits, including signed Windows status values; print the original `ExitError` for signal termination. No runtime, generated program, test selection, timeout, or scheduling changes.

Add a subprocess regression that sends SIGKILL only to itself, verifies the actual wait status, and requires the diagnostic to retain the signal. The existing success, exit-code, and launch-failure tests remain in place.

## Verification

On macOS arm64 / Go 1.27 / LLVM 22, the new regression fails against the original code (`exit code -1` instead of `signal: killed`) and passes with the fix. Focused runner tests pass with `-race -count=5`. `runNativeTest`, `runNativeTestPrograms`, `reportTestProgramResult`, and `runTestPrograms` each have 100% statement coverage in the focused run. `git diff --check` passes. CI validation of this contribution is pending.


visualfc (approved):

internal/build/run_test.go
```diff
@@ -214,6 +234,10 @@ func TestRunNativeTestHelper(t *testing.T) {
 			fmt.Fprint(os.Stderr, "stderr")
 		case "exit":
 			os.Exit(3)
+		case "signal":
+			if err := syscall.Kill(os.Getpid(), syscall.SIGKILL); err != nil {
+				t.Fatalf("kill helper: %v", err)
+			}
```
visualfc, line 240: If `Kill` returns after sending SIGKILL, the helper currently falls through to `return` and can look like a successful exit to the parent test.

Self-directed SIGKILL usually does not return, but a blocking wait (or `panic("unreachable")`) after a successful kill would make the helper deterministic:

```go
case "signal":
	if err := syscall.Kill(os.Getpid(), syscall.SIGKILL); err != nil {
		t.Fatalf("kill helper: %v", err)
	}
	select {}
```

internal/build/run.go
```diff
@@ -78,7 +78,12 @@ func runNativeTest(commands commandEnv, program testProgram, conf *Config, stdou
 	}
 	if exitErr, ok := err.(*exec.ExitError); ok {
 		if !program.coverage {
```
visualfc, line 80: Coverage programs still skip both diagnostics, so a signal-killed `llgo test -cover` run still will not print `signal: killed`.

That matches the existing contract in `TestCoverageFailedPackage` (no LLGo-specific `exit code` line). Follow-up only: the signal branch does not contain `exit code`, so it could be emitted for coverage without breaking that assertion. Out of scope here.

visualfc (approved):
## Review

The change is correct and well-scoped. `ExitError.Exited()` is the right portable check: Unix signal termination no longer reports `exit code -1`, while ordinary exits keep `exit code N`, including signed Windows status values. Printing `%v` of the `ExitError` preserves the OS reason (`signal: killed`, `signal: segmentation fault`, and `(core dumped)` when present).

The new SIGKILL helper subprocess is a good regression. `WaitStatus` confirms the signal, and matching stderr against `exitErr.Error()` already fails if the diagnostic reverts to `exit code -1`. I ran `go test -run TestRunNativeTest$ ./internal/build` and `-race -count=3` on `cbc4e0584`; both passed.

This is a diagnostic improvement only. It does not identify or fix the macOS native `test/go` crash tracked from #2578; the PR description is clear about that.

Nits below are optional and do not block merge.

internal/build/run_test.go
```diff
@@ -95,6 +97,24 @@ func TestRunNativeTest(t *testing.T) {
...
+		err := runNativeTest(commands, program, conf, io.Discard, &stderr)
+		var exitErr *exec.ExitError
+		if !errors.As(err, &exitErr) {
+			t.Fatalf("runNativeTest error = %v, want signal termination", err)
+		}
+		status, ok := exitErr.Sys().(syscall.WaitStatus)
+		if !ok || !status.Signaled() || status.Signal() != syscall.SIGKILL {
+			t.Fatalf("process status = %v, want SIGKILL", exitErr.Sys())
+		}
+		if got, want := stderr.String(), executable+": "+exitErr.Error()+"\n"; got != want {
```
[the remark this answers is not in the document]
cpunion: Updated in 8c9dbdcaa: after confirming the child was terminated by SIGKILL, the test now checks the user-facing stderr for `killed` and rejects `exit code`. This directly guards the `exit code -1` regression. `go test -race -count=3 -run '^TestRunNativeTest$' ./internal/build` passes.

internal/build/run_test.go
cpunion: Updated in 8c9dbdcaa: after a successful self-SIGKILL call, the helper now panics if it is still running, so it cannot fall through as a successful test process. The focused race test passes three consecutive runs.

internal/build/run.go
cpunion: Agreed. The coverage path retains its existing diagnostic contract, and this PR remains scoped to native non-coverage test execution. Emitting signal diagnostics for covered programs would need a separate coverage behavior/test change.

PR #2624, opened 2026-09-18
cl: preserve generic range-over-func defers

zhouguangyuan0718:
A defer inside a range-over-func loop in a generic function or method can silently disappear. Generic instances have a nonempty SSA Synthetic field, so LLGo skipped their frame when resolving the defer-stack owner and omitted the implicit RunDefers at their return.

Only treat range-over-func yield closures as transparent. Keep generic instances as real defer-owning frames and drain their explicit stack on return. Add an owner/return regression and execution coverage for generic functions and generic receiver methods, including LIFO order and captured loop values.

Validation on macOS arm64, Go 1.27, LLVM 22:

- Owner regression fails before the fix and passes afterward.
- Existing defer-owner and range-function helper tests pass.
- `llgo test -run '^TestGenericRangeFuncDefers$' -v ./test/go/rangefunc_generic_defer_test.go`: passed; before the fix both cases produced empty event lists.
- Generic receiver iterator execution matches standard Go (`3, 2, 1, b, a, ok`).

The generic-function failure also reproduces with x/tools v0.49.0; this is a pre-existing LLGo issue exposed while testing TypeScript. The generic receiver regression uses the x/tools fix from #2622, which is now merged into main.


tcka4enckoalexei2014-commits:
1

PR #2626, opened 2026-09-18
wasm: check Fiber stack bounds and expose goroutine stack size

zhouguangyuan0718:
Go tsc's AST binding exceeds a WASM worker's fixed 128 KiB stack. Without bounds checks, the write corrupts adjacent linear memory and later appears as a nil-pointer panic or incorrect TypeScript diagnostics. The same workload succeeds on the larger main-goroutine stack. Instrumented execution confirms the out-of-bounds stack-pointer update in `ast.(*Node).ForEachChild`.

Enable Emscripten's level-2 stack-pointer checks for the GoJS and Emscripten Memory32/Memory64 profiles, so overflowing a Fiber aborts before the invalid stack write. Expose the existing runtime stack budget as `-goroutine-stack-size`, retaining `-pthread-stack-size` as an alias. Deep workloads can select a suitable budget without changing default allocations for every goroutine; tsc succeeds with a 2 MiB worker stack. This change does not add dynamically growing stacks.

Add an executable recursion regression to the WASM runtime CI suite: a 128 KiB worker must report stack overflow, and the identical code must compute the correct result with a 1 MiB worker, on Memory32 and Memory64. Document the option and its Asyncify buffer cost.

Validation:
- Command-line flag tests and focused raw/named Emscripten profile tests pass.
- Executable stack-bound regressions pass for Memory32 and Memory64.
- Existing Memory32 and Memory64 scheduler fixtures pass.
- The final uninstrumented full tsc WASM build with `-goroutine-stack-size=2MB` passes `--version` and strict ES2022 compilation under Node 24.19.0. Emitted JS, cross-file generic imports and declaration files match standard-Go tsc byte for byte. TS2322 output and exit status also match; generated JavaScript runs and prints `42`.

A broader local crosscompile package run hit unrelated SDK download EOF failures; it is not counted as a passing full suite.

CI follow-up: `uintptrescapes.go` deliberately recurses 4096 levels and exceeded the default fixed goroutine stack. Its GOROOT WASM run now reserves 8 MiB, with overflow checks retained and all other cases keeping the default. The case passes locally with Go 1.26.5 sources and J32-GoJS; both J32/J64 small-stack overflow and larger-stack success probes pass. Flag aliases and GOROOT runner unit tests pass.

Browser/cache CI follow-up: replace unbounded Chrome dump-dom invocation with the existing browser acceptance driver (isolated profile, real-time guest result, 60-second process deadline); real Chrome callback passes locally. Preserve raw compiler output when cache snapshot builds fail, so a first-build setup failure cannot be hidden behind the following cache-miss mismatch. The earlier ESP32 first-build failure had its diagnostics filtered out; its cause remains unconfirmed pending the new run.


cpunion (commented):
Low priority / non-blocking: the browser check now treats the `PASS` log line as completion. The shared driver calls `finish("success")` as soon as it sees that line, so a later failure cannot change the result. I confirmed that a module which prints the expected marker and then throws still makes the driver exit 0.

The current fixture has no defer and releases its callback before printing, so this is unlikely to affect this test today. However, the WebAssembly main task and Fiber/context cleanup still run after the final `println`; a failure during that exit path (or in a defer added to the fixture later) would be missed. Could this acceptance check wait for a clean guest exit/module completion before reporting success?

cpunion (approved):

Issue #2609, opened 2026-09-16
build: funcinfo collection ignores NeedRt and leaks runtime metadata


cpunion:
## Summary

Programs with `NeedRt=false` still materialize function-information metadata from the prepared LLGo runtime tree. The runtime implementations are not present in the final executable, but their `__llgo_funcinfo_symbol_index` records are copied into the synthetic main module and increase runtime-free native binaries.

## Reproducer

`benchmark/binary_size/cprintf/main.go` only calls C `printf` through `github.com/goplus/lib/c` and does not require the LLGo runtime.

The final Linux executable confirms that the runtime implementation was not selected: it contains only the generated `runtime.main` startup wrapper and the empty weak `runtime.init` stub, and contains none of the runtime helpers whose metadata appears in the index.

Nevertheless, its function-information symbol index contains:

| Revision | Index records | Index bytes |
| --- | ---: | ---: |
| main / W1 | 796 | 12,736 |
| W2-A | 797 | 12,752 |
| W2-B / W3 | 806 | 12,896 |

Between main and W2-B, Linux `.rodata` grows by exactly 160 bytes while `.text` and the actual `.bss` section remain unchanged. The ten additional 16-byte records map to new runtime functions such as `FunctionPC`, goroutine-local initialization helpers, and panic/caller tracking helpers. Their function bodies are absent from `cprintf`; only their metadata remains.

This is an existing main issue rather than a behavior introduced by W1-W3. The original collection path and static symbol index predate those changes; the later runtime additions merely make the leak grow.

## Cause

`linkedPackageClosure` unconditionally adds the prepared runtime package tree to `linkedOrder`:

```go
visit(root)
for _, pkg := range built {
    if isRuntimePkg(pkg.PkgPath) {
        visit(pkg.Package)
    }
}
```

`planMainLink` correctly computes `needRuntime` and separates runtime archive inputs, but PCLN preparation ignores that result:

```go
funcInfo = prepareFuncInfoTableRecords(collectFuncInfo(linkedOrder), nil)
pcLineInfo = collectPCLineInfo(linkedOrder)
```

This metadata is copied directly into the generated entry module, bypassing archive-member selection and linker GC.

## Expected behavior

- With `NeedRt=false`, LLGo runtime packages should not contribute funcinfo or PC-line records to the executable.
- Metadata for the generated `runtime.main`/`runtime.goexit` frames should remain; those records are generated by the main module itself.
[...]

PR #2637, opened 2026-09-21
build: isolate runtime and cache requirements per initial program

### LINKED ISSUE #2610: build: multi-package invocation changes independent package link output
## Summary

Building several independent main packages in one `llgo build` invocation changes the link plan and binary of a package that does not use the other packages.

For example, `benchmark/binary_size/cprintf` only calls C `printf`. Building it together with `println` and `fmtprintf` makes its native executable larger and makes its Emscripten link unexpectedly include `-lembind -sDEFAULT_TO_CXX -lffi` plus many additional package archives.

This reproduces with the build cache disabled, with isolated cold and warm caches, with reversed input order, and with function information disabled. It is therefore not cache staleness, command-order nondeterminism, or only the funcinfo leak reported in #2609.

## Environment

- LLGo main: `2db247e4384848c189e94cbbe96b04462e9bc698`
- `llgo v1.0.4-0.20260915051130-2db247e43848 darwin/arm64`
- macOS 26.2 arm64
- LLVM 22.1.8
- Emscripten 6.0.8-git

## Reproducer

From the LLGo repository root, using an LLGo binary built from the revision above:

```sh
export LLGO_ROOT="$PWD"
export LLGO_BUILD_CACHE=off

llgo build -a -o /tmp/single/cprintf ./benchmark/binary_size/cprintf
llgo build -a -o /tmp/multi/ ./benchmark/binary_size/cprintf ./benchmark/binary_size/println ./benchmark/binary_size/fmtprintf

llgo build -a -target emscripten -o /tmp/wasm-single/cprintf.mjs ./benchmark/binary_size/cprintf
llgo build -a -target emscripten -o /tmp/wasm-multi/ ./benchmark/binary_size/cprintf ./benchmark/binary_size/println ./benchmark/binary_size/fmtprintf
```

Both native `cprintf` executables run and print `Hello, world`.

## Results

| Artifact | Standalone | In multi-initial build | Delta |
| --- | ---: | ---: | ---: |
| native `cprintf` | 84,480 B | 113,040 B | +28,560 B |
| Emscripten `cprintf.mjs` | 70,851 B | 82,034 B | +11,183 B |
| Emscripten `cprintf.wasm` | 113,615 B | 121,792 B | +8,177 B |

Repeating the standalone build produces the same SHA-256. Repeating the multi-initial build with the package order reversed also produces the same SHA-256, but the standalone and grouped hashes differ.

The Emscripten Wasm is structurally different, not merely wrapped by different JS:

| Section | Standalone | Grouped |
| --- | ---: | ---: |
| types | 27 | 30 |
| imports | 12 | 22 |
| functions/code entries | 177 | 198 |
| exports | 26 | 30 |
| code bytes | 104,912 | 110,381 |

The final standalone Emscripten link does not contain `-lembind`, `-sDEFAULT_TO_CXX`, or `-lffi`. The final link for the exact same `cprintf` package in the grouped build contains all three and substantially more archive inputs.

### Independent of funcinfo

With `LLGO_FUNCINFO=off`, the problem remains:

| Artifact | Standalone | In multi-initial build | Delta |
| --- | ---: | ---: | ---: |
| native `cprintf` | 33,600 B | 78,880 B | +45,280 B |
| Emscripten `cprintf.mjs` | 70,851 B | 82,034 B | +11,183 B |
| Emscripten `cprintf.wasm` | 113,615 B | 121,792 B | +8,177 B |

This issue is related to #2609 because both expose an overly broad `linkedOrder`, but it is broader: real archives, link arguments, imports, and function bodies leak between independent initials.

### Cache behavior

`NeedRt` is currently derived per package. It is deliberately not part of the package fingerprint, but is stored in manifest metadata and restored on cache hits:

- save: `metadata.need_rt` in `saveToCache`
- restore: `pkg.NeedRt = meta.NeedRt` in `tryLoadFromCache`

I tested standalone and grouped builds with caching enabled, then repeated both as warm builds. The warm standalone build had 14 package hits and one miss; the warm grouped build had 78 hits and six misses (main packages are intentionally not cached). Repeated artifacts were byte-identical within each scenario. Correction: the original test used copied compiler binaries with distinct timestamps; a clean, versioned compiler does not necessarily include executable metadata in its compiler hash, so that method did not prove empty cache namespaces. The September 21 follow-up below uses an explicit unique compiler-hash namespace and verifies zero cold hits.

| Artifact | Standalone repeated SHA-256 | Grouped repeated SHA-256 |
| --- | --- | --- |
| `cprintf.mjs` | `d83b680e53d0c585c97e5d383f178d13ab564776d0562b95d7ec5aee56fc14f0` | `ec54940da22e5c1d6efed829281fe796ab6f85e30e66d8b238a8cd5fcbb2e485` |
| `cprintf.wasm` | `1412946b196f2e0d7f59a1ec994b627741eb63966937a77daad7e442d4fb0ae7` | `c5c22b9aaab6fd44513dc260a9e29b056f39e63e0d51e3b45474387c02a6bdd5` |

The cache preserves the selected behavior correctly; it is not the source of the cross-initial contamination.

## Cause

There are two intended levels of runtime selection:

1. `buildAllPkgs` aggregates `packageRuntimeNeeds(normalTasks)` across the invocation to decide whether the shared target runtime packages need to be prepared at all.
2. `planMainLink` recomputes `needRuntime` for each individual main package before selecting runtime archives.

Sharing runtime compilation work at level 1 is reasonable, but the final per-initial isolation is lost. `linkedPackageClosure` starts from one root and then unconditionally visits every prepared runtime package in the invocation:

```go
visit(root)
for _, pkg := range built {
    if isRuntimePkg(pkg.PkgPath) {
        visit(pkg.Package)
    }
}
```

`visit` recursively follows each runtime package's effective dependencies. In `planMainLink`, only paths recognized by `isRuntimePkg` are deferred as runtime inputs. Dependencies outside that path are handled as ordinary packages: their `NeedRt`, `NeedPyInit`, `LinkArgs`, snapshots, and archives all affect the current root. Thus a runtime tree prepared because of one initial can make another initial require the runtime and can inject unrelated provider flags such as libffi/embind.

For native builds, all runtime tasks are prepared up front, and the same global-built-set closure also makes an initial's result depend on which other roots were loaded in the invocation.

## Expected behavior

- Compiling several independent main packages together must produce the same per-package link plan as compiling each package separately.
- Compilation/backend work may be shared globally, but semantic dependency closure, `NeedRt`, provider `LinkArgs`, archives, package init lists, reflection data, and PCLN metadata must be selected per initial.
- A runtime-free `cprintf` must not acquire libffi/embind because another initial needs them.
- Cache-off, cold-cache, and warm-cache builds must make the same selection.

## Suggested regression coverage

1. Build a runtime-free C program alone and together with runtime-using programs; compare its native and Emscripten link plans and final artifacts.
2. Repeat with cache disabled, isolated cold and warm caches, and reversed initial order.
3. Assert that grouped Emscripten `cprintf` does not acquire `-lembind`, `-sDEFAULT_TO_CXX`, or `-lffi`.
4. Add a focused cache round-trip test for `NeedRt`, `NeedPyInit`, and `LinkArgs` metadata.
5. Cover multi-package `ModeTest` as well, because it uses the same package build/link planner.

cpunion:
## Summary

Fixes #2610.

`NeedRt` is already recorded per package and restored from its manifest. The contamination came from two other places: each link included every prepared runtime-tree overlay, and a shared compilation used invocation-wide local-context/WASM feature decisions for unrelated initial programs.

- Compute each executable's link closure from its own ordinary/alternate imports plus the core runtime and its dependencies, not unrelated runtime overlays prepared for other initials.
- Compute local-context startup, WASI reflection bridges, and WASM function-entry metadata requirements per initial. Only initials with identical code-generation settings share a backend group. Ordinary same-configuration batches retain shared compilation and the native test DAG; incompatible groups use the existing worker budget sequentially.
- Retain the existing feature-sensitive cache keys. This isolates incompatible package variants while letting standalone and grouped builds reuse the same compatible archives, without changing `NeedRt` serialization or adding global cache invalidation.
- Preserve multi-build output/error semantics, test-package paths, and the parent build trace across groups.
- Release the parent's analysis/LLVM resources before building children. Prepare graph-free child invocations, copy feature values without retaining their package group, and group native inputs before unnecessary union SSA construction. WASM retains the required per-initial SSA analysis.

## Regression coverage

- Compare actual cprintf/println/fmtprintf executables, package fingerprints, and runtime requirements across standalone/grouped builds, reversed initial order, cold/hot caches, and both cache population orders. Require byte identity within each cache population and retain fingerprint/runtime/execution checks across independent cold populations, whose dependency-init call order and ELF layout padding can differ even in standalone builds. On Windows, compare size/fingerprints/execution within each population rather than timestamp-bearing PE bytes.
- Cover unrelated overlay dependencies, per-initial locality and WASM settings, compatible batching, check-only output, independent link failures, shared build tracing, and generated test-package names.
- Add Emscripten cprintf single/grouped/warm-single byte comparisons to the existing WebAssembly profile smoke job, with no additional CI matrix.
- Verify with a weak reference that prepared child invocations do not retain the parent's package graph.

## Local validation

- Full `go test ./internal/build ./ssa -count=1 -timeout=15m`: passed in the initial validation (build 89.0%, SSA 80.9% package coverage).
- After the memory-lifetime follow-up, the complete `internal/build` suite passes again with 89.0% package coverage; `Build`, `groupInitialBuilds`, `initialGroupInvocations`, and `buildInitialGroups` have 100% statement coverage in that run.
- Native macOS arm64 and Emscripten wasm32 cprintf: standalone and grouped outputs are byte-identical.
- `dev/test_wasm_target_profiles.sh` passes after the lifetime follow-up for Emscripten wasm32/wasm64, `GOOS=js GOARCH=wasm`, WASI target and `GOOS=wasip1 GOARCH=wasm`, including the artifact checks. An instrumented compiler confirms that the actual WASM child-plan handoff executes.

## Memory investigation

On macOS arm64 / Go 1.27 / LLVM 22 with GOMAXPROCS=16, a cold mixed cprintf/println/fmtprintf build peaked at 1,127.58 MiB before the lifetime follow-up (median of three runs), versus 797.23 MiB after it (one run); the pinned main control was 786.53 MiB (median of three). RSS is sampled for the process tree, can double-count shared pages, and is not Linux cgroup usage.

The full-suite cumulative leak had a separate cause: main's global Plan 9 signature cache retained completed build contexts. That independent fix was merged through #2641 and is now present in this PR's base; this PR does not duplicate it. Before that merge, the grouping-only full suite still peaked at about 7.16 GiB, so the localized mixed-build improvement must not be mistaken for the cumulative-cache reduction. Local measurements do not prove that earlier runner disconnections were OOM.

The runtime-free native metadata issue #2609 was handled independently by the now-merged #2636. This PR is rebased onto main containing #2636 and #2641; its own diff remains limited to multi-initial isolation, regression coverage, and the associated analysis-lifetime reduction.


internal/build/initial_groups.go
```diff
@@ -0,0 +1,77 @@
...
+	"errors"
+	"strings"
+
+	"github.com/xgo-dev/llgo/internal/packages"
+)
+
+// These settings affect package code, not just the final link. Independent
+// programs may share a backend/cache entry only when these settings agree.
+type initialBuildFeatures struct {
+	localContext    bool
```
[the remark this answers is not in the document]
cpunion: Addressed in 70f318ae5: documented that localContext groups compatible initials but is re-derived from each child package subset; the two WASM decisions are passed explicitly.

internal/build/initial_groups.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	var groups []initialBuildGroup
+	indexes := make(map[initialBuildFeatures]int)
+	target := ctx.prog.Target()
+	for _, pkg := range ctx.initial {
+		features := initialBuildFeatures{
+			localContext: ctx.prog.NeedsLocalContextForPackages(activeLocalityPackages([]*packages.Package{pkg}, alts)),
+		}
+		if target.GOARCH == "wasm" {
+			roots := wasmReflectRoots(&context{progSSA: ctx.progSSA, initial: []*packages.Package{pkg}})
```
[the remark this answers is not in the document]
cpunion: Addressed in 70f318ae5: cache the conservative unrooted scan within groupInitialBuilds while retaining separate rooted RTA analyses. Added coverage for multiple rootless initials across all three providers.

internal/build/initial_groups.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, group := range groups {
+		conf := ctx.buildConf.clone()
+		args := make([]string, len(group.pkgs))
+		for i, pkg := range group.pkgs {
+			args[i] = pkg.PkgPath
+			if ctx.mode == ModeTest {
+				args[i] = strings.TrimSuffix(args[i], ".test")
+			}
+		}
+		child := inv
```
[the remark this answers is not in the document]
cpunion: Addressed in 70f318ae5: documented preservation of invocation controls and why single-root fallback calls cannot enter grouping. Also fixed the discovered check-only singleton-group output-path issue and added existing-file/directory preservation coverage.

internal/build/invocation.go
```diff
@@ -23,6 +23,11 @@ type Invocation struct {
 	// caller-visible ModeBuild still has normal cmd/go output semantics.
 	compileOnly          bool
 	disableMultiFallback bool
+	// Feature groups preserve the original multi-build output contract while
+	// compiling only initials that can safely share package artifacts.
+	multipleInitials bool
+	initialFeatures  *initialBuildFeatures
+	parentBuildTrace *buildTracer
```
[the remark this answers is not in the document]
cpunion: Addressed in 70f318ae5: documented the borrowed tracer ownership on parentBuildTrace. The shared trace regression continues to verify parent and child build spans.

cpunion:
Follow-up 70f318ae5 also documents the deliberate frontend rebuild per incompatible group. The end-to-end WASM child-Build path is exercised by dev/test_wasm_target_profiles.sh (single/grouped/warm-single Emscripten cprintf comparisons), already wired into the existing wasm-runtime profile smoke job. Local _xtool build ./... and the strengthened check-only output preservation regression now pass; no test skip or coverage threshold relaxation was added.

cpunion:
CI update for 70f318ae5: Windows (MSVC/MinGW), macOS, LLGo/WASM integration and Dev LTO now pass; codecov/patch passes at 97.73%. The remaining Ubuntu Go job has twice ended with the GitHub annotation “The self-hosted runner lost communication with the server”, with no retrievable job log or reported test assertion.

There is a possible runner lifecycle/mapping issue: in attempt 2, Ubuntu job 106221739335 ran on e2b-106221739132, while Dev LTO job 106221739132 ran on e2b-106221739335. Dev LTO completed at 05:27:21 UTC; Ubuntu was marked disconnected at 05:45:55 UTC. The first attempt similarly used a runner named after another job and was marked disconnected about 18 minutes after that other job completed. This is evidence for infrastructure investigation, not a confirmed root cause.

Attempt 3 reruns only Ubuntu, preserving all successful jobs and avoiding simultaneous startup with Dev LTO. No tests, assertions, or coverage thresholds were disabled.

Attempt 3 used the matching runner and completed normally. It exposed an overbroad assertion comparing fmtprintf bytes across independent cold cache populations. A local standalone-only experiment reproduced the same-size, same-fingerprint difference: internal/sync.init swapped the calls to sync/atomic.init and internal/race.init, with the corresponding Mach-O UUID/signature changes. Commit b8e395718 now compares artifact bytes within each cache population, while retaining cross-population fingerprints, both population directions, reversed roots, sizes and execution checks. Five consecutive isolation/cache regression runs pass (137.96 s), followed by three more runs (78.68 s) with an explicit cross-population size assertion. The production implementation is unchanged; CI is running on the updated head.

Infrastructure follow-up: Ubuntu job 106236200281 on the updated head also lost communication despite a matching runner name (e2b-106236200281). Therefore the name-swapping observation does not establish the root cause. Its logs are again unavailable (404). The previous isolated Ubuntu run did complete normally; after the other jobs finish, only Ubuntu will be rerun.

The isolated retry on b8e395718 (job 106247104081, matching runner e2b-106247104081) also ended with the lost-communication annotation, and its log endpoint returns 404. All other checks pass, including Windows/macOS/WASM and codecov/patch (97.73%). No compiler assertion failure is available from these disconnected runs. Repeating the unchanged job is no longer diagnostic; runner-side lifecycle/resource/network logs or validation on an alternative runner are needed. No coverage threshold or test skip has been changed.

cpunion:
Rebased onto current main at 9da890955, which includes the merged #2636 and #2641. The PR still contains four multi-initial isolation commits and no duplicate metadata/cache-lifetime changes. Post-rebase validation passed: focused internal/build grouping, cache, error/output and Plan 9 cache regressions; SSA locality tests; and the complete WebAssembly target profile script covering Emscripten wasm32/wasm64, raw GoJS, target/raw WASI, plus single/grouped/warm-single artifact equality.

cpunion:
CI failure diagnosis and fix: Ubuntu failed only in TestMultiBuildRuntimeIsolationAndCache because two independent cold cache populations produced fmtprintf ELF files 8 bytes apart (1,639,808 vs 1,639,800). The package fingerprint, runtime requirement and execution checks agreed; independent cold builds already permit different dependency-init order, and ELF layout/alignment can therefore differ slightly as well. Commit c58281f0f removes only that cross-population exact-size assertion. Artifact equality (or Windows size equality) remains required within each population, and all semantic/cache checks remain required across populations. Three local repeats pass. The codecov/patch failure was secondary: the Ubuntu test step exited before uploading the primary coverage profile, leaving only partial reports.

internal/build/initial_groups.go
```diff
@@ -0,0 +1,110 @@
...
+	return groups
+}
+
+func initialGroupInvocations(inv Invocation, ctx *context, groups []initialBuildGroup) []Invocation {
+	children := make([]Invocation, 0, len(groups))
+	for _, group := range groups {
+		conf := ctx.buildConf.clone()
+		args := make([]string, len(group.pkgs))
+		for i, pkg := range group.pkgs {
+			args[i] = pkg.PkgPath
```
visualfc: Child groups reload by `pkg.PkgPath` rather than a user-loadable pattern. That works inside a module (and the new tests all use modules or in-repo paths), but GOPATH / no-`go.mod` initials often have `PkgPath` `_/abs/path`, which `packages.Load` / `go list` will not accept.

If two such mains differ in `localContext`, the parent splits and the child `Build` fails; the same command succeeds when they stay in one group. Prefer a loadable pattern (`filepath.Rel(dir, pkg.Dir)`, or map back to `inv.Args`) so grouping cannot change whether the packages are found.

internal/build/initial_groups.go
```diff
@@ -0,0 +1,110 @@
...
+			localContext: ctx.prog.NeedsLocalContextForPackages(activeLocalityPackages([]*packages.Package{pkg}, alts)),
+		}
+		if target.GOARCH == "wasm" {
+			roots := wasmReflectRoots(&context{progSSA: ctx.progSSA, initial: []*packages.Package{pkg}})
+			var use *wasmProgramUse
+			if len(roots) == 0 {
+				// Library-style roots use the same conservative function set.
+				// Unlike rooted RTA, this scan can be shared across initials.
+				if unrootedUse == nil {
+					unrootedUse = analyzeWasmProgramUse(ctx.progSSA, nil)
```
visualfc: `analyzeWasmProgramUse(ctx.progSSA, nil)` is `ssautil.AllFunctions` over the **union** SSA. Sharing that scan means a library-style initial inherits another initial's `reflect.Call` / `FuncForPC` uses. `TestGroupInitialWasmFeatures` currently asserts the unrooted packages join the reflecting group, so this is locked in.

Rooted mains (the #2610 case) are fine because they use per-initial RTA. `llgo install pkg1 pkg2` and other entry-less wasm batches still share features. Please document that this path is intentionally conservative, or scan only functions belonging to that initial.

internal/build/initial_groups_test.go
```diff
@@ -0,0 +1,265 @@
...
+		conf.BuildParallelism = 7
+		return children, weak.Make(pkg)
+	}()
+	child := children[0]
+	if !slices.Equal(child.Args, []string{"example.com/first"}) || child.Dir != "source" ||
+		child.Config.BuildParallelism != 3 || !child.initialFeatures.localContext ||
+		!child.disableMultiFallback || !child.multipleInitials {
+		t.Fatalf("group invocation lost its independent snapshot: %+v", child)
+	}
+	runtime.GC()
```
visualfc: A single `runtime.GC()` does not guarantee the unreachable `packages.Package` is collected (race detector, high GOGC, incomplete sweep). This can fail closed and report a retain that is not there.

Retry until `parent.Value()` is nil, for example:

```go
for i := 0; i < 10 && parent.Value() != nil; i++ {
    runtime.GC()
}
```

Copying `features := group.features` before taking the address is the right retain fix; only the assertion here is brittle.

ssa/locality.go
```diff
@@ -394,26 +394,42 @@ func (p Program) PackageLocalitiesFor(pkg *types.Package) map[string]VariableLoc
...
+	defer p.localities.mu.RUnlock()
+	return p.localities.needsContext(active)
+}
+
+func (p *localityInfos) needsContext(activePackages map[string]struct{}) bool {
 	needsContext := func(info VariableLocality) bool {
 		return info.Locality != locality.None && (info.LocalStorage != LocalStorageNativeTLS || hasInitialization(info.Info))
 	}
-	for _, info := range p.localities.ownerlessEntries {
+	for _, info := range p.ownerlessEntries {
```
visualfc: `declarationEntries` is filtered by the supplied package set, but `ownerlessEntries` and the leftover `entries` loop are not. The parent has already parsed every initial before grouping, so one ownerless/legacy entry that needs a context forces **all** initials to `localContext=true`. They then share a group, and a program that would not need LocalContext standalone still gets the startup path.

Production parse mostly uses `DeclareLocality` / `SetLocalityInfoFor`, so this may be empty today, but it contradicts the per-program API. Please ignore ownerless/undeclared entries unless they belong to the queried packages, and add a test that an ownerless declaration on package A does not make `NeedsLocalContextForPackages(B)` true.

visualfc (commented):
Review of the remaining isolation/lifetime issues. The main #2610 path (per-initial link closure, feature-homogeneous groups, check-only temp outputs, borrowed trace ownership) looks correct. Please address the four inline notes before merging.

visualfc (approved):

internal/build/initial_groups.go
cpunion: Addressed in 55bf667bb. Child groups now keep normal import paths, but translate GOPATH-style _/ paths and command-line-arguments to a loadable path relative to the original invocation directory using pkg.Dir. The regression verifies ./first reloads through packages.Load with module mode disabled; the existing synthetic .test package round-trip remains covered.

internal/build/initial_groups.go
cpunion: Addressed in 55bf667bb. WASM feature analysis is now per initial: executable packages use RTA from init/main, while entry-less packages start RTA from functions owned by that package. An initial absent from the SSA graph receives an empty analysis instead of the union-wide AllFunctions result. The test now checks both main and library initials across WASI, GoJS, and Emscripten and verifies an unrelated initial remains feature-free.

internal/build/initial_groups_test.go
cpunion: Addressed in 55bf667bb. The weak-reference assertion now retries collection for up to ten GC cycles before reporting retained parent package state. The focused test also passes under the race detector.

ssa/locality.go
cpunion: Addressed in 55bf667bb. Per-package locality queries now filter ownerless and legacy undeclared entries by the selected canonical package paths, while declaration-owned entries remain filtered by concrete owner. The legacy all-package NeedsLocalContext behavior is preserved. Added coverage for selected/unrelated/empty package sets, ownerless metadata, legacy metadata, and nil package entries.

tcka4enckoalexei2014-commits:
1

